PR #15, opened 2016-12-07
pull updated UDD and Reporter, propagate errors

DwayneJengSage:
Changes:
* pulled in bug fixes from UDD and Reporter
* fix version conflict with GSON that was preventing runtime
* remove unused old Java SDK
* propagate errors back to the PollSqsWorker - This has the side effect of making Worker Platform single-threaded.

Testing done (both UDD and Reporter):
* injected errors, verified message is cycled back to SQS
* Bad Request errors, verified these are *not* cycled back to SQS
* normal case

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>Bridge-Reporter</artifactId>
-            <version>1.0.1</version>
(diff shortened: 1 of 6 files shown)


liujoshua (approved):

zaclinsage (approved):
LGTM

alxdarksage:
Looks fine to me as well

PR #23, opened 2017-11-15
Switch to using new Bridge Platform worker queue

zaro0508:
The infra for BridgeWorkerPlatform app is now automated with
BridgeWorkerPlatform-infra[1] repo.  This project uses cloudformation
to setup the EB worker env and SQS queues. We should switch to using
this queue and deprecate the manually created SQS queues.

[1] https://github.com/Sage-Bionetworks/BridgeWorkerPlatform-infra


DwayneJengSage (approved):
You'll also need to update the queues in DDB table Reporter-Scheduler-Config.

Note that this config may or may not be going away when we implement a smarter scheduler, which I think we'll be tackling probably in January.

zaro0508:
Reporter-Scheduler-Config DDB tables have been updated with new sqs queues

PR #26, opened 2017-11-29
update JavaSDK

DwayneJengSage:


DwayneJengSage:
Merged as part of emergency Worker fix

PR #30, opened 2017-12-24
BRIDGE-2024: monitor with new relic

zaro0508:
Setup agent to allow EC2 monitoring with new relic.
Note: only works when deployed to EB bridgeworker-* Envs

### DIFF
--- a/ebextensions/newrelic.config
+++ b/ebextensions/newrelic.config
@@ -1,0 +1,24 @@
+files:
+  "/etc/newrelic-infra.yml" :
+    mode: "000644"
+    owner: root
+    group: root
+    content: |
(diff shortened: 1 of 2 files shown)


DwayneJengSage (commented):
What happens when this is deployed to the old WorkerPlatform envs?

zaro0508:
It will work in the old WorkerPlatform env the only requirement is that  'NEW_RELIC_LICENSE_KEY' & 'NEW_RELIC_APP_NAME' EB env vars must be setup prior to the deployment.  Otherwise the agent will start but just fail to connect to NR and report errors to the newrelic-infra.log file.

DwayneJengSage (approved):

PR #35, opened 2018-01-12
BRIDGE-2024: Stream tomcat log to cloudwatch

zaro0508:
BridgeWorker runs in a tomcat process and its application log is sent
to the tomcat log.  Use cloudwatch log integration[1] to stream the
application logs to cloudwatch so we can monitor messages in the log.

[1] https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/AWSHowTo.cloudwatchlogs.html

### DIFF
--- a/ebextensions/tomcatlogs.config
+++ b/ebextensions/tomcatlogs.config
@@ -1,0 +1,32 @@
+###################################################################################################
+#### The configuration below sets the logs to be pushed, the Log Group name to push the logs to and
+#### the Log Stream name as the instance id. The following files are examples of logs that will be
+#### streamed to CloudWatch Logs in near real time:
+#### 
+#### /var/log/tomcat8/catalina.out
+#### 
+#### You can then access the CloudWatch Logs by accessing the AWS CloudWatch Console and clicking
+#### the "Logs" link on the left. The Log Group name will follow this format:
+####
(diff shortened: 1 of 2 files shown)


DwayneJengSage (commented):
Why do we need a special .ebextensions file just to stream logs? How do we handle this in BridgePF and why can't we do the same thing here?

zaro0508:
By default, EB streams different logs depending on the AMI that is used.  Goto the link in the commit message and you'll see the default logs for different AMIs.

For BridgePF we use "java" AMI and EB configures the following logs:
- /var/log/eb-activity.log
- /var/log/nginx/access.log
- /var/log/nginx/error.log
- /var/log/web-1.error.log
- /var/log/web-1.log

For BridgeWorker we use "tomcat" AMI and EB configures these logs:
- /var/log/eb-activity.log
- /var/log/httpd/error_log
- /var/log/httpd/access_log
- /var/log/nginx/error_log
- /var/log/nginx/access_log

For any other logs you need to do this setup, basically tell the cloudwatch agent to configure other logs.  This is how it works as I understand it from reading the documentation.  If you prefer not do this setup then you could probably refactor the BridgeWorker app to write to one of the default logs that is setup by EB (in above tomcat list).

DwayneJengSage (approved):

PR #42, opened 2018-04-24
So that all worker processes run in the worker app, some dependencies have to be advanced.

alxdarksage:
This should probably be merged before UDD version updates.


DwayneJengSage:
Depends on https://github.com/Sage-Bionetworks/BridgeUserDataDownloadService/pull/24

Also, how was this tested?

DwayneJengSage (approved):

PR #48, opened 2018-07-14
randomized messages and decouple data group logic

DwayneJengSage:
JIRAs:
* https://sagebionetworks.jira.com/browse/BRIDGE-2173 - Notification Worker should randomize from a list of messages
* https://sagebionetworks.jira.com/browse/BRIDGE-2235 - Missed Activities notification should not depend on data groups
* ~~https://sagebionetworks.jira.com/browse/BRIDGE-2256 - Pre-burst SMS notification should be on the first day of the burst~~

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/notification/worker/BridgeNotificationWorkerProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/notification/worker/BridgeNotificationWorkerProcessor.java
@@ -5,6 +5,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -50,6 +51,7 @@ public class BridgeNotificationWorkerProcessor implements ThrowingConsumer<JsonN
 
     private BridgeHelper bridgeHelper;
     private DynamoHelper dynamoHelper;
+    private Random rng = new Random();
 
     /** Bridge helper. */
     @Autowired
@@ -66,6 +68,11 @@ public class BridgeNotificationWorkerProcessor implements ThrowingConsumer<JsonN
     /** Set rate limit, in users per second. This is primarily to allow unit tests to run without being throttled. */
     public final void setPerUserRateLimit(double rate) {
         perUserRateLimiter.setRate(rate);
+    }
+
+    /** Allows mocking of the RNG. */
+    public void setRng(Random rng) {
+        this.rng = rng;
     }
 
     /** Main entry point into the Notification Worker. */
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/notification/worker/WorkerConfig.java
```diff
@@ -82,48 +83,45 @@ public void setExcludedDataGroupSet(Set<String> excludedDataGroupSet) {
...
     }
 
-    /** @see #getMissedCumulativeActivitiesMessagesByDataGroup */
-    public void setMissedCumulativeActivitiesMessagesByDataGroup(
-            Map<String, String> missedCumulativeActivitiesMessagesByDataGroup) {
-        this.missedCumulativeActivitiesMessagesByDataGroup = missedCumulativeActivitiesMessagesByDataGroup != null ?
-                ImmutableMap.copyOf(missedCumulativeActivitiesMessagesByDataGroup) : ImmutableMap.of();
+    /** @see #getMissedCumulativeActivitiesMessagesList */
+    public void setMissedCumulativeActivitiesMessagesList(List<String> missedCumulativeActivitiesMessagesList) {
+        this.missedCumulativeActivitiesMessagesList = missedCumulativeActivitiesMessagesList != null ?
```
alxdarksage, line 95: As this isn't initialized to null, would this be clearer/shorter:
if (missedCumulativeActivitiesMessagesList != null) {
    this. missedCumulativeActivitiesMessagesList = missedCumulativeActivitiesMessagesList;
}

src/main/java/org/sagebionetworks/bridge/notification/worker/BridgeNotificationWorkerProcessor.java
```diff
@@ -410,36 +410,39 @@ private void notifyUser(String studyId, StudyParticipant participant, Notificati
...
-            }
-        }
-        if (message == null) {
+        if (messageList == null) {
             throw new IllegalStateException("No messages found for type " + notificationType + " for user " + userId);
         }
 
+        // Pick message at random.
+        int randomIndex = rng.nextInt(messageList.size());
+        String message = messageList.get(randomIndex);
```
alxdarksage, line 444: As someone who makes a lot of generators, I've learned that pure randomness sometimes produces very annoying results, like the same message 12 times. Rotating through the messages per user might be better for the user, but requires having some state so probably isn't worth it here.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/notification/worker/WorkerConfig.java
DwayneJengSage: As per IM conversation, the semantics of this setter are that (a) null values are replaced with empty lists and (b) we copy the list so that changes in the original list don't get reflected through. I think this is the cleanest way to do this.

PR #55, opened 2018-09-24
Redrive Worker Multithreading

DwayneJengSage:
We need to redrive 7k uploads per day for BP Lab. Synchronous redrive takes ~5 seconds, because of a server-side polling loop, so 7k uploads takes ~9.2 hours.

We'll want to speed this up by adding multithreading to the worker.

Testing done:
* mvn verify (unit tests, FindBugs, Jacoco test coverage)
* Redrive integ tests
* Manually tested UDD (the other worker that uses thread pools)


alxdarksage (approved):
Looks good

PR #56, opened 2018-09-26
Redrive Worker - Move upload polling to client-side

DwayneJengSage:
This moves the upload polling from server-side to client-side to reduce the number of long requests the server has to deal with. This helps reduce the number of socket timeouts.

Testing done:
* mvn verify (unit tests, FindBugs, Jacoco coverage)
* integ tests


alxdarksage (approved):
Does this mean we should remove the code to poll on the server?

PR #66, opened 2019-04-25
FitBit worker should not retry infinitely on users

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2055

Added an error limit so that the FitBit worker doesn't retry infinitely on users.

Testing done:
* Hacked up my local server to throw a 500 on OAuthController.getAccessToken() and tested manually.
* mvn verify (unit tests, FindBugs, Jacoco test coverage)
* integ tests

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/fitbit/worker/BridgeFitBitWorkerProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/fitbit/worker/BridgeFitBitWorkerProcessor.java
@@ -26,12 +26,14 @@ import org.sagebionetworks.bridge.fitbit.util.Utils;
 import org.sagebionetworks.bridge.rest.model.Study;
 import org.sagebionetworks.bridge.sqs.PollSqsWorkerBadRequestException;
 import org.sagebionetworks.bridge.worker.ThrowingConsumer;
+import org.sagebionetworks.bridge.workerPlatform.exceptions.WorkerException;
 
 /** Worker consumer for the FitBit Worker. This is called by BridgeWorkerPlatform and is the main entry point. */
 @Component("FitBitWorker")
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/fitbit/worker/BridgeFitBitWorkerProcessor.java
```diff
@@ -178,6 +187,15 @@ void processStudy(String dateString, Study study) {
...
                     LOG.error("Error getting next user: " + ex.getMessage(), ex);
+
+                    // The Iterator is a paginated iterator that calls Bridge for each user. If for some reason, it
+                    // keeps throwing exceptions (for example, Bridge is down), this could retry infinitely. Cap the
+                    // number of errors, and if we hit that threshold, break out of the loop and propagate the
+                    // exception up the call stack.
+                    numErrors++;
+                    if (numErrors >= userErrorLimit) {
+                        throw new WorkerException("User error limit reached, aborting for study " + studyId);
+                    }
```
alxdarksage, line 198: Could possibly use the generic LimitExceededException here.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/fitbit/worker/BridgeFitBitWorkerProcessor.java
DwayneJengSage: I'd prefer to avoid mixing Bridge Server exceptions and Bridge Worker exceptions.

alxdarksage: It's a generic exception in the Java standard libraries.

PR #71, opened 2019-06-26
FitBit Sleep Data

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2519

Requires https://github.com/Sage-Bionetworks/bridge-base/pull/51


src/main/resources/schema.json
```diff
@@ -180,5 +180,102 @@
...
+                        "columnId":"infoCode",
+                        "columnType":"INTEGER"
+                    },
+                    {
+                        "columnId":"isMainSleep",
+                        "columnType":"BOOLEAN"
+                    },
+                    {
+                        "columnId":"levels",
+                        "columnType":"LARGETEXT"
```
DanWebster, line 215: Does LARGETEXT in this case capture all of the nested structures within levels?  For example, would this column be populated with the following:
`{
                "summary": {
                    "deep": {
                        "count": <value>,
                        "minutes": <value>,
                        "thirtyDayAvgMinutes": <value>
                    },
                    "light": {
                        "count": <value>,
                        "minutes": <value>,
                        "thirtyDayAvgMinutes": <value>
                    },
                    "rem": {
                        "count": <value>,
                        "minutes": <value>,
                        "thirtyDayAvgMinutes": <value>
                    },
                    "wake": {
                        "count": <value>,
                        "minutes": <value>,
                        "thirtyDayAvgMinutes": <value>
                    }
                }`

alxdarksage (approved):

src/main/resources/schema.json
DwayneJengSage: Yes, a LARGETEXT would just copy the entire JSON blob into the Synapse table column.

PR #72, opened 2019-07-02
FitBit Raw Data

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2520

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/fitbit/worker/UserProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/fitbit/worker/UserProcessor.java
@@ -252,7 +252,7 @@ public class UserProcessor {
     private String uploadJsonAsFileHandle(RequestContext ctx, JsonNode node, String filename) throws IOException,
             SynapseException {
         // Append a random string to the filename to make it probabilistically unique.
-        String uniqueFilename = filename + RandomStringUtils.randomAlphabetic(4);
+        String uniqueFilename = filename + RandomStringUtils.randomAlphabetic(4) + ".json";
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/fitbit/worker/UserProcessor.java
```diff
@@ -254,6 +246,30 @@ String serializeJsonForColumn(RequestContext ctx, JsonNode node, ColumnSchema co
         return String.valueOf(value);
     }
 
+    // Helper method for uploading the given JSON node as a Synapse file handle. This method also appends a random
+    // 4-character alphabetical string to the filename to minimize the probability of a file collision. This method
+    // both creates and deletes the temporary file.
+    private String uploadJsonAsFileHandle(RequestContext ctx, JsonNode node, String filename) throws IOException,
+            SynapseException {
+        // Append a random string to the filename to make it probabilistically unique.
+        String uniqueFilename = filename + RandomStringUtils.randomAlphabetic(4);
```
alxdarksage: Does the filename ever have an extension? Would you want to insert like this instead: "file.zip" --> "file-<random>.zip" ?

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/fitbit/worker/UserProcessor.java
DwayneJengSage: These are actually JSON files, and the filenames (currently based on JSON keys) don't contain any file extension. So I added the .json file extension.

PR #77, opened 2019-08-19
Retention Reports

janaechan:
Retention reports when the user last signed in and uploaded data in relation to the days in the study

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
@@ -2,40 +2,25 @@ package org.sagebionetworks.bridge.reporter.helper;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-import org.joda.time.Days;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import com.google.common.collect.HashMultiset;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Multiset;
-
+import org.sagebionetworks.bridge.notification.helper.AccountSummaryIterator;
 import org.sagebionetworks.bridge.reporter.worker.Report;
 import org.sagebionetworks.bridge.rest.ClientManager;
-import org.sagebionetworks.bridge.rest.api.AuthenticationApi;
-import org.sagebionetworks.bridge.rest.api.ForAdminsApi;
 import org.sagebionetworks.bridge.rest.api.ForWorkersApi;
-import org.sagebionetworks.bridge.rest.api.HealthDataApi;
 import org.sagebionetworks.bridge.rest.api.StudiesApi;
 import org.sagebionetworks.bridge.rest.model.AccountSummary;
 import org.sagebionetworks.bridge.rest.model.AccountSummaryList;
-import org.sagebionetworks.bridge.rest.model.ActivityEvent;
 import org.sagebionetworks.bridge.rest.model.ActivityEventList;
-import org.sagebionetworks.bridge.rest.model.HealthDataRecord;
-import org.sagebionetworks.bridge.rest.model.HealthDataSubmission;
 import org.sagebionetworks.bridge.rest.model.ReportData;
 import org.sagebionetworks.bridge.rest.model.RequestInfo;
-import org.sagebionetworks.bridge.rest.model.ResourceList;
-import org.sagebionetworks.bridge.rest.model.SignIn;
-import org.sagebionetworks.bridge.rest.model.SignUp;
 import org.sagebionetworks.bridge.rest.model.Study;
 import org.sagebionetworks.bridge.rest.model.StudyParticipant;
 import org.sagebionetworks.bridge.rest.model.Upload;
@@ -117,58 +102,29 @@ public class BridgeHelper {
         return retList;
     }
     
-    public Map<StudyParticipant, RequestInfo> getRequestInfoForParticipant (String studyId, List<StudyParticipant> participants)
-            throws IOException {
-    	Map<StudyParticipant, RequestInfo> retMap = new HashMap<>();
-        
+    public StudyParticipant getStudyPartcipant (String studyId, String userId) throws IOException {
         ForWorkersApi workersApi = bridgeClientManager.getClient(ForWorkersApi.class);
-        for (StudyParticipant participant : participants) {
-        	RequestInfo requestInfo = workersApi.getRequestInfoForWorker(studyId, participant.getId()).execute().body();
-    		retMap.put(participant, requestInfo);
-    		doSleep();
-        }
-        return retMap;
+        return workersApi.getParticipantByIdForStudy(studyId, userId, false).execute().body();
     }
     
-    public Map<RequestInfo, ActivityEventList> getActivityEventForParticipant (String studyId, List<StudyParticipant> participants)
-            throws IOException {
-    	Map<RequestInfo, ActivityEventList> retMap = new HashMap<>();
-        
+    public RequestInfo getRequestInfoForParticipant (String studyId, String userId) throws IOException {
         ForWorkersApi workersApi = bridgeClientManager.getClient(ForWorkersApi.class);
-        for (StudyParticipant participant : participants) {
-        	RequestInfo requestInfo = workersApi.getRequestInfoForWorker(studyId, participant.getId()).execute().body();
-        	ActivityEventList activityEvents = workersApi.getActivityEventsForParticipantAndStudy(studyId, participant.getId()).execute().body();
-    		retMap.put(requestInfo, activityEvents);
-    		doSleep();
-        }
-        return retMap;
+        return workersApi.getRequestInfoForWorker(studyId, userId).execute().body();
     }
     
-    public void uploadHealthData() throws IOException {
-    	String APP_VERSION = "version 1.0.0, build 2";
-        DateTimeZone CREATED_ON_TIMEZONE = DateTimeZone.forOffsetHours(9);
-        String PHONE_INFO = "Integration Tests";
-        String SCHEMA_ID = "health-data-integ-test-schema";
-        long SCHEMA_REV = 1L;
-        Map<String, String> data = ImmutableMap.<String, String>builder().put("answer-me", "C").build();
-        DateTime createdOn = DateTime.now(CREATED_ON_TIMEZONE);
-//        DateTime createdOn = new DateTime(2019, 8, 25, 9, 30);
-        HealthDataApi healthDataApi = bridgeClientManager.getClient(HealthDataApi.class);
-        
-        HealthDataSubmission submission = new HealthDataSubmission().appVersion(APP_VERSION).createdOn(createdOn)
-                .data(data).phoneInfo(PHONE_INFO).schemaId(SCHEMA_ID).schemaRevision(SCHEMA_REV);
-        
-        healthDataApi.submitHealthData(submission).execute().body();        
+    public ActivityEventList getActivityEventForParticipant (String studyId, String userId) throws IOException {
+        ForWorkersApi workersApi = bridgeClientManager.getClient(ForWorkersApi.class);
+        return workersApi.getActivityEventsForParticipantAndStudy(studyId, userId).execute().body();
     }
     
-    public void signUpandSignIn(String study_id, String email) throws IOException {
-        AuthenticationApi authenticationApi = bridgeClientManager.getClient(AuthenticationApi.class);
-    	SignUp signUp = new SignUp().study(study_id).email(email).password("P@ssword`1");
-        SignIn signIn = new SignIn().study(study_id).email(email).password("P@ssword`1");
+    /**
+     * Get an iterator for all account summaries in the given study. Note that since getAllAccountSummaries is a
+     * paginated API, the iterator may continue to call the server.
+     */
+    public Iterator<AccountSummary> getAllAccountSummaries(String studyId) {
+        return new AccountSummaryIterator(bridgeClientManager, studyId);
+    }
 
-        authenticationApi.signUp(signUp).execute().body();
-        authenticationApi.signIn(signIn).execute().body();
-    }
     
     private void doSleep() {
         // sleep a second
--- a/src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
@@ -2,21 +2,40 @@ package org.sagebionetworks.bridge.reporter.helper;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.Days;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Multiset;
+
 import org.sagebionetworks.bridge.reporter.worker.Report;
 import org.sagebionetworks.bridge.rest.ClientManager;
+import org.sagebionetworks.bridge.rest.api.AuthenticationApi;
+import org.sagebionetworks.bridge.rest.api.ForAdminsApi;
 import org.sagebionetworks.bridge.rest.api.ForWorkersApi;
+import org.sagebionetworks.bridge.rest.api.HealthDataApi;
 import org.sagebionetworks.bridge.rest.api.StudiesApi;
 import org.sagebionetworks.bridge.rest.model.AccountSummary;
 import org.sagebionetworks.bridge.rest.model.AccountSummaryList;
+import org.sagebionetworks.bridge.rest.model.ActivityEvent;
+import org.sagebionetworks.bridge.rest.model.ActivityEventList;
+import org.sagebionetworks.bridge.rest.model.HealthDataRecord;
+import org.sagebionetworks.bridge.rest.model.HealthDataSubmission;
 import org.sagebionetworks.bridge.rest.model.ReportData;
+import org.sagebionetworks.bridge.rest.model.RequestInfo;
+import org.sagebionetworks.bridge.rest.model.ResourceList;
+import org.sagebionetworks.bridge.rest.model.SignIn;
+import org.sagebionetworks.bridge.rest.model.SignUp;
 import org.sagebionetworks.bridge.rest.model.Study;
 import org.sagebionetworks.bridge.rest.model.StudyParticipant;
 import org.sagebionetworks.bridge.rest.model.Upload;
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
```diff
@@ -98,6 +117,59 @@ public final void setBridgeClientManager(ClientManager bridgeClientManager) {
...
+        ForWorkersApi workersApi = bridgeClientManager.getClient(ForWorkersApi.class);
+        for (StudyParticipant participant : participants) {
+        	RequestInfo requestInfo = workersApi.getRequestInfoForWorker(studyId, participant.getId()).execute().body();
+    		retMap.put(participant, requestInfo);
+    		doSleep();
+        }
+        return retMap;
+    }
+    
+    public Map<RequestInfo, ActivityEventList> getActivityEventForParticipant (String studyId, List<StudyParticipant> participants)
```
alxdarksage: I would key this map by something like user ID. I'm not 100% sure that two RequestInfo objects generated by the Swagger codegen libraries won't have the same hashCode (value equality). User ID is guaranteed to be unique per user.

src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
```diff
@@ -98,6 +117,59 @@ public final void setBridgeClientManager(ClientManager bridgeClientManager) {
...
+        Map<String, String> data = ImmutableMap.<String, String>builder().put("answer-me", "C").build();
+        DateTime createdOn = DateTime.now(CREATED_ON_TIMEZONE);
+//        DateTime createdOn = new DateTime(2019, 8, 25, 9, 30);
+        HealthDataApi healthDataApi = bridgeClientManager.getClient(HealthDataApi.class);
+        
+        HealthDataSubmission submission = new HealthDataSubmission().appVersion(APP_VERSION).createdOn(createdOn)
+                .data(data).phoneInfo(PHONE_INFO).schemaId(SCHEMA_ID).schemaRevision(SCHEMA_REV);
+        
+        healthDataApi.submitHealthData(submission).execute().body();        
+    }
```
alxdarksage, line 117: This looks like test data in the BridgeHelper. This should be moved to a test class only in the test class path.

src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
```diff
@@ -98,6 +117,59 @@ public final void setBridgeClientManager(ClientManager bridgeClientManager) {
...
+    
+    public void signUpandSignIn(String study_id, String email) throws IOException {
+        AuthenticationApi authenticationApi = bridgeClientManager.getClient(AuthenticationApi.class);
+    	SignUp signUp = new SignUp().study(study_id).email(email).password("P@ssword`1");
+        SignIn signIn = new SignIn().study(study_id).email(email).password("P@ssword`1");
+
+        authenticationApi.signUp(signUp).execute().body();
+        authenticationApi.signIn(signIn).execute().body();
+    }
```
alxdarksage, line 125: Same. This looks like test data, it should be under src/test not src/main.

src/main/java/org/sagebionetworks/bridge/reporter/request/ReportType.java
```diff
@@ -5,7 +5,8 @@
 public enum ReportType {
     DAILY("DAILY"), // aka DAILY_UPLOADS 
     WEEKLY("WEEKLY"), // aka WEEKLY_UPLOADS
-    DAILY_SIGNUPS("DAILY_SIGNUPS");
+    DAILY_SIGNUPS("DAILY_SIGNUPS"),
+	DAILY_RETENTION("DAILY_RETENTION");
```
alxdarksage: Replace tab with four spaces. Four spaces per indentation level is the standard for all Bridge java code.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public Report generate(BridgeReporterRequest request, String studyId) throws IOException {
+    	
+        DateTime startDate = request.getStartDateTime();
+        DateTime endDate = request.getEndDateTime();
+        String scheduler = request.getScheduler();
+        ReportType scheduleType = request.getScheduleType();
+        String reportId = scheduler + scheduleType.getSuffix();
+        
+        List<StudyParticipant> participants = bridgeHelper.getParticipantsForStudy(studyId, startDate, endDate);
```
alxdarksage: I'm a little worried about loading all participants and their request info objects or a study with many users, like Cardiovascular, which has 63,000 users. Iterating through the users and processing each in turn might be better.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        
+        // Total number of participants: last_sign_in vs days_in_study
+        Map<StudyParticipant, RequestInfo> requestInfos = bridgeHelper.getRequestInfoForParticipant(studyId, participants);
+        List<Integer> signInData = new ArrayList<>();
+        for (Entry<StudyParticipant, RequestInfo> entry : requestInfos.entrySet()) {
+        	if (entry.getValue().getSignedInOn() != null) { // Has signed in
+        		System.out.println("Participant: " + entry.getKey().getEmail());
+        		System.out.println("Created on: " + entry.getKey().getCreatedOn().toLocalDate() + " Signed in on: " + entry.getValue().getSignedInOn().toLocalDate());
```
alxdarksage: Remove sop statements in your final code. :-)

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        DateTime startDate = request.getStartDateTime();
+        DateTime endDate = request.getEndDateTime();
+        String scheduler = request.getScheduler();
+        ReportType scheduleType = request.getScheduleType();
+        String reportId = scheduler + scheduleType.getSuffix();
+        
+        List<StudyParticipant> participants = bridgeHelper.getParticipantsForStudy(studyId, startDate, endDate);
+        
+        // Total number of participants: last_sign_in vs days_in_study
+        Map<StudyParticipant, RequestInfo> requestInfos = bridgeHelper.getRequestInfoForParticipant(studyId, participants);
```
DwayneJengSage: I wouldn't recommend using a mutable object as a hash key. It's really easy to accidentally modify the hash key, and then suddenly your entry disappears from the hashtable.

I'd recommend instead creating a map from String to RequestInfo, where the key is StudyParticipant.getId().

You'd also need to create a second map from String to StudyParticipant. Guava Maps.uniqueIndex https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Maps.html#uniqueIndex-java.lang.Iterable-com.google.common.base.Function- should be able to do this for you.

```
Maps.uniqueIndex(participants, StudyParticipant::getId);
```

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        List<Integer> signInData = new ArrayList<>();
+        for (Entry<StudyParticipant, RequestInfo> entry : requestInfos.entrySet()) {
+        	if (entry.getValue().getSignedInOn() != null) { // Has signed in
+        		System.out.println("Participant: " + entry.getKey().getEmail());
+        		System.out.println("Created on: " + entry.getKey().getCreatedOn().toLocalDate() + " Signed in on: " + entry.getValue().getSignedInOn().toLocalDate());
+        		int days = Days.daysBetween(entry.getKey().getCreatedOn().toLocalDate(), entry.getValue().getSignedInOn().toLocalDate()).getDays();
+        		System.out.println("Days between: " + days);
+        		while (signInData.size() < (days + 1)) {
+            		signInData.add(0);
+            	}
```
DwayneJengSage: Looks like there's mixed tabs and spaces here. I'd recommend double checking Eclipse's code style settings and making sure we're converting tabs into 4 spaces and using 4-space indent.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        		signInData.set(days, signInData.get(days) + 1);
+        	}
+        }
+        System.out.println("Sign In Data: " + signInData.toString());
+        
+        // Total number of participants: last_uploaded_on vs days_in_study
+        Map<RequestInfo, ActivityEventList> activityEventList = bridgeHelper.getActivityEventForParticipant(studyId, participants);
```
DwayneJengSage: Similarly, you'll want to map participant ID to ActivityEventList. Plus, there's no reason to load RequestInfo a second time since we already have a map of RequestInfos.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        List<Integer> uploadedOnData = new ArrayList<>();
+        for (Entry<RequestInfo, ActivityEventList> entry : activityEventList.entrySet()) {
+        	if (entry.getKey().getSignedInOn() != null) { // Has signed in
+        		for (int i = 0; i < entry.getValue().getItems().size(); i++) {
+    	    		if (entry.getValue().getItems().get(i).getEventId().equals("activities_retrieved")) {
+    	    			System.out.println("Activity Retrieved  on: " + entry.getValue().getItems().get(i).getTimestamp().toDateTime().toLocalDate() + " Uploaded on: " + entry.getKey().getUploadedOn().toLocalDate());
+    	    			int days = Days.daysBetween(entry.getValue().getItems().get(i).getTimestamp().toDateTime().toLocalDate(), entry.getKey().getUploadedOn().toLocalDate()).getDays();
```
DwayneJengSage: These are some very long lines. I'd recommend setting a max line width to 120 characters to make the code easier to read. IntelliJ helpfully shows a purple line (color configurable) where 120 characters is. I dunno if Eclipse has a similar setting.

DwayneJengSage: There seems to be some inconsistency here. You check if RequestInfo.getSignedInOn() is null, but you actually use getUploadedOn. Also, for bySignIn, you index to participant.getCreatedOn(), but here you index to activities_retrieved. These should probably use the same index for consistency.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+    	        		System.out.println("Days between: " + days);
+    	        		while (uploadedOnData.size() < (days + 1)) {
+    	    				uploadedOnData.add(0);
+    	            	}
+    	    			uploadedOnData.set(days, uploadedOnData.get(days) + 1);
+    	    		}
+        		}
```
DwayneJengSage: A participant only has one activities_retrieved event. Once you find it, you should break out of the loop.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    
+    @Override
+    public Report generate(BridgeReporterRequest request, String studyId) throws IOException {
+    	
+        DateTime startDate = request.getStartDateTime();
+        DateTime endDate = request.getEndDateTime();
+        String scheduler = request.getScheduler();
+        ReportType scheduleType = request.getScheduleType();
+        String reportId = scheduler + scheduleType.getSuffix();
```
DwayneJengSage: I know this is the same pattern we use elsewhere, but I don't think it's necessary to include the scheduler in the report name. The ScheduleType already uniquely identifies the report, and the scheduler is generally a long string (which is not used anywhere else) and we end up with a long unreadable report name.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
DwayneJengSage: I agree with Alx here. Fortunately, we have an AccountSummaryIterator that does exactly what you want https://github.com/Sage-Bionetworks/BridgeWorkerPlatform/blob/develop/src/main/java/org/sagebionetworks/bridge/notification/helper/BridgeHelper.java#L38 Note that this lives in the Notification Worker's BridgeHelper, so you'll need to copy it into the Reporter's BridgeHelper. (We have a task to refactor all the BridgeHelpers into a single BridgeHelper.

Once you have an iterator of Account Summaries, you can then load the Study Participant, Request Info, and Activity Events by User ID.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
janaechan: Eclipse isn't applying the max line width setting, so I might need to go in and shorten lines manually

janaechan:
Should I add a before() method with the annotation @BeforeMethod in BridgeHelperTest to set up the ForWorkerApi, ClientManager, and BridgeHelper to remove some redundancy?

src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
```diff
@@ -98,6 +102,30 @@ public final void setBridgeClientManager(ClientManager bridgeClientManager) {
...
+        return workersApi.getActivityEventsForParticipantAndStudy(studyId, userId).execute().body();
+    }
+    
+    /**
+     * Get an iterator for all account summaries in the given study. Note that since getAllAccountSummaries is a
+     * paginated API, the iterator may continue to call the server.
+     */
+    public Iterator<AccountSummary> getAllAccountSummaries(String studyId) {
+        return new AccountSummaryIterator(bridgeClientManager, studyId);
+    }
```
DwayneJengSage, line 125: Did you have to make a change to AccountSummaryIterator to get this to work?

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+        Iterator<AccountSummary> accountSummaryIter = bridgeHelper.getAllAccountSummaries(studyId);
+
+        List<Integer> signInData = new ArrayList<>();
+        List<Integer> uploadedOnData = new ArrayList<>();
+        while (accountSummaryIter.hasNext()) {
+            AccountSummary accountSummary = accountSummaryIter.next();
+            StudyParticipant studyParticipant = bridgeHelper.getStudyPartcipant(studyId, accountSummary.getId());
+            RequestInfo requestInfo = bridgeHelper.getRequestInfoForParticipant(studyId, accountSummary.getId());
+            ActivityEventList activityEventList = bridgeHelper.getActivityEventForParticipant(studyId, accountSummary.getId());
+            DateTime studyStartDate = studyParticipant.getCreatedOn();
```
DwayneJengSage: If we always re-calculate study_start_date on the server-side, do we still need to fetch the Participant to get the createdOn?

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+        String scheduler = request.getScheduler();
+        ReportType scheduleType = request.getScheduleType();
+        String reportId = scheduleType.getSuffix();
+
+        Iterator<AccountSummary> accountSummaryIter = bridgeHelper.getAllAccountSummaries(studyId);
+
+        List<Integer> signInData = new ArrayList<>();
+        List<Integer> uploadedOnData = new ArrayList<>();
+        while (accountSummaryIter.hasNext()) {
+            AccountSummary accountSummary = accountSummaryIter.next();
```
DwayneJengSage: I'd recommend wrapping the inside of the loop in a try-catch block. Otherwise, if a single participant has an error, the whole report errors out.

Secondly, I'd recommend using a RateLimiter of some sort to limit how many requests you are sending at once. There are 250k accounts in prod right now, so if we RateLimit to 100 accounts per second, we'll finish the report in 45 minutes. (Currently discussing with Alx about whether we should use a whitelist to only run this report on selected studies, but that would happen in the scheduler, not the worker.)

Similarly, we should increase the page_size in AccountSummaryIterator to 100, which is the max.

DwayneJengSage (commented):
> Should I add a before() method with the annotation @BeforeMethod in BridgeHelperTest to set up the ForWorkerApi, ClientManager, and BridgeHelper to remove some redundancy?

I would recommend doing so, yes. Removing repeated code makes maintaining and updating easier.

src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
janaechan: No changes were made to the code. I created a AccountSummaryIterator (copied from notification.helper) in reporter.helper but this call to AccountSummaryIterator still routes to the notification.helper iterator.

DwayneJengSage: Doesn't the one from notification.helper only show accounts with phone numbers?

src/test/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelperTest.java
```diff
@@ -189,6 +195,86 @@ public void testSaveReportForStudy() throws Exception {
...
+        ClientManager mockClientManager = mock(ClientManager.class);
+        when(mockClientManager.getClient(ForWorkersApi.class)).thenReturn(mockWorkerClient);
+        ActivityEventList mockActiviyEventList = mockCallForActivityEventList(mockWorkerClient, USER_ID_1);
+        
+        BridgeHelper bridgeHelper = new BridgeHelper();
+        bridgeHelper.setBridgeClientManager(mockClientManager);
+        ActivityEventList activityEventList = bridgeHelper.getActivityEventForParticipant(TEST_STUDY_ID, USER_ID_1);
+        
+        assertNull(activityEventList.getItems());
+        assertEquals(activityEventList.getItems(), mockActiviyEventList.getItems());
```
DwayneJengSage: You might want to try assertSame() to verify the activityEventList you're getting from the workerApi is the same activityEventList being returned by the BridgeHelper. Right here, you're just comparing null to null. (assertSame() might also be better for the other tests as well.)

src/test/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGeneratorTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.sagebionetworks.bridge.rest.model.AccountSummary;
+import org.sagebionetworks.bridge.rest.model.ActivityEvent;
+import org.sagebionetworks.bridge.rest.model.ActivityEventList;
+import org.sagebionetworks.bridge.rest.model.RequestInfo;
+import org.sagebionetworks.bridge.reporter.helper.BridgeHelper;
+import org.sagebionetworks.bridge.reporter.request.ReportType;
+import org.sagebionetworks.bridge.rest.model.StudyParticipant;
+import org.testng.annotations.Test;
+
+public class RetentionReportGeneratorTest {
```
DwayneJengSage, line 30: Maybe add a test with two or three users to make sure the report is being aggregated correctly.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelper.java
janaechan: Oh right! I forgot I made that change, thanks for remembering!

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
janaechan: getSuffix() is only used when generating reports. Can I remove the leading "-" of getSuffix() and remove the scheduler from from the report name?

src/test/java/org/sagebionetworks/bridge/reporter/helper/BridgeHelperTest.java
```diff
@@ -190,10 +197,67 @@ public void testSaveReportForStudy() throws Exception {
...
+    @Test
+    public void testGetActivityEventForParticipant() throws Exception {
+        when(mockClientManager.getClient(ForWorkersApi.class)).thenReturn(mockWorkerClient);
+        ActivityEventList mockActiviyEventList = mockCallForActivityEventList(mockWorkerClient, USER_ID_1);
+        
+        ActivityEventList activityEventList = bridgeHelper.getActivityEventForParticipant(TEST_STUDY_ID, USER_ID_1);
         
-        ClientManager mockClientManager = mock(ClientManager.class);
+        assertNull(activityEventList.getItems());
+        assertSame(activityEventList.getItems(), mockActiviyEventList.getItems());
```
DwayneJengSage: You should call assertSame on the activityEventList itself. Otherwise, you're comparing null to null, which is trivially true.

DwayneJengSage (changes requested):
Looks good. Just one comment on the assert.

Similar to Server2, you should run `mvn verify` on this. Also, you should run `mvn spring-boot:run` on both Server2 and Worker and run your updated WorkerTest against your Worker. (Note that FitBitWorker will fail. That's expected, as that requires a lot of specific setup. The reporter tests should work though.)

DwayneJengSage (changes requested):
Aren't we also pulling in a new API for getRequestInfo? There should be an update to pom.xml as well.

DwayneJengSage (commented):
Looks good. Looks like there's a merge conflict you need to resolve.

Also, do we have changes for BridgeWorkerIntegrationTests?

DwayneJengSage (approved):

PR #82, opened 2019-09-05
Signup and Retention Report bug fixes

janaechan:
https://sagebionetworks.jira.com/browse/BRIDGE-2555
https://sagebionetworks.jira.com/browse/BRIDGE-2556
https://sagebionetworks.jira.com/browse/BRIDGE-2557

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
+++ b/src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
@@ -83,8 +83,12 @@ public class RetentionReportGenerator implements ReportGenerator {
                 RequestInfo requestInfo = bridgeHelper.getRequestInfoForParticipant(
                         studyId, accountSummary.getId());
                 if (requestInfo.getSignedInOn() != null) {
-                    int sign_in_days = Days.daysBetween(studyStartDate, 
+                    int sign_in_days = Days.daysBetween(studyStartDate.withZone(DateTimeZone.UTC), 
                             requestInfo.getSignedInOn().withZone(DateTimeZone.UTC)).getDays();
+                    if (sign_in_days < 0) {
+                        LOG.error("study_state_date is negative for id=" + accountSummary.getId());
+                        continue;
+                    }
                     
                     while (signInData.size() < (sign_in_days + 1)) {
                         signInData.add(0);
@@ -92,8 +96,12 @@ public class RetentionReportGenerator implements ReportGenerator {
                     signInData.set(sign_in_days, signInData.get(sign_in_days) + 1);
                 }
                 if (requestInfo.getUploadedOn() != null) {
-                    int upload_on_days = Days.daysBetween(studyStartDate, 
+                    int upload_on_days = Days.daysBetween(studyStartDate.withZone(DateTimeZone.UTC), 
                             requestInfo.getUploadedOn().withZone(DateTimeZone.UTC)).getDays();
+                    if (upload_on_days < 0) {
+                        LOG.error("upload_on_days is negative for id=" + accountSummary.getId());
+                        continue;
+                    }
                     
                     while (uploadedOnData.size() < (upload_on_days + 1)) {
                         uploadedOnData.add(0);
--- a/src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
+++ b/src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.joda.time.Days;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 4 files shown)


src/test/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGeneratorTest.java
```diff
@@ -51,7 +56,7 @@
     RetentionReportGenerator generator;
     
     @Mock
-    private BridgeHelper bridgeHelper;
+    private static BridgeHelper bridgeHelper;
```
DwayneJengSage: I don't think we want to make this static. I'm not sure how Mockito handles this, and I think it would be safer to make it non-static, to ensure it's not re-used across test cases.

src/test/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGeneratorTest.java
```diff
@@ -38,8 +40,11 @@
     
     private static final DateTime STUDY_START_DATE_1 = DateTime.parse("2017-06-05T00:00:00.000Z");
     private static final DateTime STUDY_START_DATE_2 = DateTime.parse("2017-06-04T00:00:00.000Z");
+    
     private static final DateTime SIGN_IN_ON = DateTime.parse("2017-06-09T00:00:00.000Z");
-    private static final DateTime UPLOADED_ON = DateTime.parse("2017-06-09T00:00:00.000Z");
+    private static final DateTime SIGN_IN_ON_TIMEZONE = DateTime.parse("2017-06-04T06:50:21.650-07:00");
```
DwayneJengSage: We should pick a time late at night, like 18:00-0700, to make sure that converting the timezone makes a difference.

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -65,20 +74,27 @@ public Report generate(BridgeReporterRequest request, String studyId) throws IOE
...
                     continue;
                 }
-                RequestInfo requestInfo = bridgeHelper.getRequestInfoForParticipant(studyId, accountSummary.getId());
+                
+                RequestInfo requestInfo = bridgeHelper.getRequestInfoForParticipant(
+                        studyId, accountSummary.getId());
                 if (requestInfo.getSignedInOn() != null) {
-                    int sign_in_days = Days.daysBetween(studyStartDate.toLocalDate(), requestInfo.getSignedInOn().toLocalDate()).getDays();
+                    int sign_in_days = Days.daysBetween(studyStartDate, 
+                            requestInfo.getSignedInOn().withZone(DateTimeZone.UTC)).getDays();
```
DwayneJengSage, line 87: Even though studyStartDate is always in UTC, we should explicitly set the time zone here, just to be safe. (Also when we calculate upload_on_days.)

src/main/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGenerator.java
```diff
@@ -65,20 +74,27 @@ public Report generate(BridgeReporterRequest request, String studyId) throws IOE
...
+                    
                     while (signInData.size() < (sign_in_days + 1)) {
```
DwayneJengSage, line 93: Just to be extra safe, we should check if sign_in_days is somehow negative, and then continue on to the next participant if it is. (Same with upload_on_days.)

src/test/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGeneratorTest.java
```diff
@@ -82,9 +89,41 @@ public void testStudyStartDate() throws Exception {
...
+        List<AccountSummary> accountSummaries = new ArrayList<>();
+        accountSummaries.add(accountSummary);
+        Iterator<AccountSummary> accountSummaryIter = accountSummaries.iterator();
+        when(bridgeHelper.getAllAccountSummaries(STUDY_ID)).thenReturn(accountSummaryIter);
+        
+        mockStudyParticipant(bridgeHelper, USER_ID_1, new ArrayList<>());
+        
+        mockActivityEventList(accountSummary, mockStudyStateDateEvent(STUDY_START_DATE_1));
+        
+        RequestInfo requestInfo = mockRequestInfo(SIGN_IN_ON_TIMEZONE, UPLOADED_ON);
```
DwayneJengSage: We should also have UPLOADED_ON with timezone, to make sure that we have both cases covered.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/reporter/worker/RetentionReportGeneratorTest.java
```diff
@@ -38,8 +40,14 @@
     
     private static final DateTime STUDY_START_DATE_1 = DateTime.parse("2017-06-05T00:00:00.000Z");
     private static final DateTime STUDY_START_DATE_2 = DateTime.parse("2017-06-04T00:00:00.000Z");
+    
     private static final DateTime SIGN_IN_ON = DateTime.parse("2017-06-09T00:00:00.000Z");
-    private static final DateTime UPLOADED_ON = DateTime.parse("2017-06-09T00:00:00.000Z");
+    private static final DateTime SIGN_IN_ON_INVALID = DateTime.parse("2016-06-09T00:00:00.000Z");
+    private static final DateTime SIGN_IN_ON_TIMEZONE = DateTime.parse("2017-06-04T06:50:21.650-07:00");
```
DwayneJengSage: 6:50-0700 is 13:50Z. We should pick something later in the day that will actually change the date to verify that our timestamp logic is working correctly.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #84, opened 2019-10-02
Consolidate BridgeHelpers, DynamoHelpers, Exceptions, and Constants

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2584

Testing done:
* mvn verify (unit tests, Jacoco test coverage, FindBugs)
* integ tests
* manually tested UserDataDownload (which can't be easily integ tested)


src/main/java/org/sagebionetworks/bridge/udd/helper/SnsHelper.java
```diff
@@ -2,9 +2,9 @@
 
 import java.util.Map;
 
-import org.sagebionetworks.bridge.udd.accounts.AccountInfo;
-import org.sagebionetworks.bridge.udd.dynamodb.StudyInfo;
 import org.sagebionetworks.bridge.udd.s3.PresignedUrlInfo;
+import org.sagebionetworks.bridge.workerPlatform.bridge.AccountInfo;
+import org.sagebionetworks.bridge.workerPlatform.dynamodb.StudyInfo;
```
alxdarksage, line 7: I don't know about a package name that has a capital in it... I've never seen that before. In terms of the principle of least surprise, I'd expect it to be workerplatform.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/AccountSummaryIterator.java
```diff
@@ -38,12 +40,14 @@ public AccountSummaryIterator(ClientManager clientManager, String studyId) {
     private void loadNextPage() {
         // Call server for the next page.
         try {
-            // The offset into the next page is equal to the number of accounts that we have seen.
             // HACK: We use "1" for the phone filter. This is because we want to filter out any accounts that don't
             // have phone numbers. Right now, all phone accounts are in the US, so we can simply use the country code
             // ("1") in the filter.
```
alxdarksage: FWIW there are some Indian phone numbers in the system.

alxdarksage (approved):
Just one comment about workerPlatform as a package name (I have never seen camel case in a package name before).

src/main/java/org/sagebionetworks/bridge/udd/helper/SnsHelper.java
DwayneJengSage: If we were creating a new package, I'd be inclined to change it to workerplatform (or maybe just worker). However, the package already exists and has a bunch of files in it, and I don't think we should rename it at this point.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/AccountSummaryIterator.java
DwayneJengSage: Updated the comment. We only support sending to US phone numbers. (Or at least the code as written will only send to phone numbers in the North America Numbering Plan. If we want to support mPower 2.0 to send reminder notifications to non-US numbers, that's a separate discussion and a separate code change.

PR #94, opened 2020-05-26
Update bridge-base to relax deser constraints on unknown properties

alxdarksage:


alxdarksage:
Does not have all the changes necessary to build.

PR #95, opened 2020-05-26
Minimum changes required to update default ObjectMapper in bridge-base

alxdarksage:
See https://sagebionetworks.jira.com/browse/BRIDGE-2844

In updating the default object mapper, we pull in changes to the Synapse client and the UploadSchemaKey, but these are the only changes needed so it'll compile and pass tests. This should allow requests to succeed until we change the BridgeWorkerPlatform to use appId instead of study.


DwayneJengSage (commented):
It looks like your WorkerPlatform codebase is very very old. Can you merge the latest changes and re-submit the pull request?

DwayneJengSage (approved):

PR #101, opened 2021-03-16
Download participant roster worker 

kllysng:
Here's a rough first working version of the participant roster worker processor. I'm really looking forward to your feedback. Thank you!


src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+        try {
+            request = DefaultObjectMapper.INSTANCE.treeToValue(jsonNode, BridgeDownloadParticipantRosterRequest.class);
+        } catch (IOException e) {
+            throw new PollSqsWorkerBadRequestException("Error parsing request: " + e.getMessage(), e);
+        }
+
+        String userId = request.getUserId();
+        String appId = request.getAppId();
+        String password = request.getPassword();
+        LOG.info("Received request for userId=" + userId + ", app=" + appId + ", password=" + password);
```
DwayneJengSage: It's okay to put the userId and appId in the logs, but not the password. If the logs somehow get leaked, this password would allow someone to potentially steal the participant roster.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+        File zipFile = null;
+
+        try {
+            // get caller's user using Bridge API getParticipantByIdForApp
+            StudyParticipant participant = bridgeHelper.getParticipant(appId, userId, false);
+            String orgMembership = participant.getOrgMembership();
+
+            // call Bridge API searchAccountSummariesForApp(appId, caller's Org)
+            int offsetBy = 0;
+            List<AccountSummary> accountSummaries = bridgeHelper.getAccountSummariesForApp(appId, orgMembership, offsetBy);
```
DwayneJengSage: Should we pass in the page size into bridgeHelper.getAccountSummariesForApp()? This way, if PAGE_SIZE in this class somehow becomes different from PAGE_SIZE in BridgeHelper, we don't end up with weird bugs.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+
+            csvFile = fileHelper.newFile(tmpDir, getDownloadFilenamePrefix() + ".csv");
+            try (CSVWriter csvFileWriter = new CSVWriter(fileHelper.getWriter(csvFile))) { //TODO is this nested try catch block an anti pattern? :/
```
DwayneJengSage: Nested try-catch block is fine. The outer try-catch block does general error handling and ensures we clean up after the worker task. Inner try-catch block handles File IO. In this particular case, this is actually a try-with-resources, which opens the FileWriter and automatically closes it for us when we're done.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+                    Thread.sleep(THREAD_SLEEP_INTERVAL);
+                }
+            } catch (IOException ex) {
+                throw new WorkerException("Error creating file " + csvFile + ": " + ex.getMessage(), ex);
+            }
+
+            // zip up and email the csv file
+            AppInfo appInfo = dynamoHelper.getApp(appId);
+            AccountInfo accountInfo = bridgeHelper.getAccountInfo(appId, userId);
+            String zipFileName = "userdata-" + UUID.randomUUID().toString() + ".zip";
```
DwayneJengSage: The random UUID is not required for the Participant Roster Worker. It was only required for User Data Download because UDD uses S3, and we didn't want an S3 filename collision. It's fine for the Participant Roster Worker because we use a temp dir, and the temp dir is randomly generated.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+            fileHelper.deleteFile(zipFile);
+            fileHelper.deleteDir(tmpDir);
+            LOG.info("request took " + requestStopwatch.elapsed(TimeUnit.SECONDS) +
+                    " seconds for userId =" + userId + ", app=" + appId + ", password=" + password);
+        }
+    }
+
+    protected String getDownloadFilenamePrefix() {
+        return "account_summaries";
+    }
```
DwayneJengSage: Is this modeled after something? Do we need to make a method to return a string? Or can we make it a static final variable?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
DwayneJengSage: Do we need to write the headers for the CSV?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private String[] getAccountSummaryArray(AccountSummary accountSummary) {
+        JsonElement json = RestUtils.toJSON(accountSummary);
+
+        Set<Map.Entry<String, JsonElement>> members = json.getAsJsonObject().entrySet();
+        ArrayList<String> list = new ArrayList<>();
+
+        for (Map.Entry<String, JsonElement> member : members) {
+            list.add(member.getValue().toString());
+        }
```
DwayneJengSage: Do we want to get a list of column names, so we can guarantee that we always go through the map entries in a consistent order?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+
+        Stopwatch requestStopwatch = Stopwatch.createStarted();
+        File csvFile = null;
+        File tmpDir = fileHelper.createTempDir();
+        File zipFile = null;
+
+        try {
+            // get caller's user using Bridge API getParticipantByIdForApp
+            StudyParticipant participant = bridgeHelper.getParticipant(appId, userId, false);
+            String orgMembership = participant.getOrgMembership();
```
DwayneJengSage: Should probably add a quick sanity check here that the participant has an email address. This will ensure we don't spend a lot of work downloading a roster just to throw because we have a null email address.

Also, we should check that emailVerified is true. This way, we don't accidentally spam somebody who didn't verify their email address.

If both of those checks fail, the cleanest thing to do is log a warning and return.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (BridgeSDKException ex) {
+            int status = ex.getStatusCode();
+            if (status >= 400 && status < 500) {
+                throw new PollSqsWorkerBadRequestException(ex);
+            } else {
+                throw new RuntimeException(ex);
+            }
+        } finally {
+            fileHelper.deleteFile(csvFile);
+            fileHelper.deleteFile(zipFile);
```
DwayneJengSage: We should check that these files are not null before we try to delete them.

src/main/java/org/sagebionetworks/bridge/udd/helper/SesHelper.java
```diff
@@ -123,4 +178,54 @@ private void sendEmailToAccount(AppInfo appInfo, AccountInfo accountInfo, Body b
...
+
+        message.setContent(mimeMessage);
+
+        mimeMessage.addBodyPart(wrap);
+
+        // define attachment
+        MimeBodyPart mimeAttachment = new MimeBodyPart();
+        DataSource dataSource = new FileDataSource(attachment);
+        mimeAttachment.setDataHandler(new DataHandler(dataSource));
+        mimeAttachment.setFileName(dataSource.getName());
```
DwayneJengSage, line 234: Do we need to set mime type "text/csv"?

src/main/java/org/sagebionetworks/bridge/workerPlatform/config/SpringConfig.java
```diff
@@ -66,6 +67,7 @@
         "org.sagebionetworks.bridge.udd",
         "org.sagebionetworks.bridge.uploadredrive",
         "org.sagebionetworks.bridge.workerPlatform",
+        "org.sagebionetworks.bridge.participantRoster" //TODO not sure I need this one or the bean below
```
DwayneJengSage: This is correct, because we want to do ComponentScan on the whole package. The SpringConfig files listed below are for older legacy workers.

DwayneJengSage (changes requested):
The overall structure is good. Just a few comments.

src/main/java/org/sagebionetworks/bridge/udd/helper/SesHelper.java
kllysng: Do you mean we will want to set it as type "application/zip"?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,231 @@
...
+
+        Stopwatch requestStopwatch = Stopwatch.createStarted();
+        File csvFile = null;
+        File tmpDir = fileHelper.createTempDir();
+        File zipFile = null;
+
+        try {
+            // get caller's user
+            StudyParticipant participant = bridgeHelper.getParticipant(appId, userId, false);
+            String orgMembership = participant.getOrgMembership();
```
DwayneJengSage, line 127: After chatting with Alx, we need another check in the worker. We need to check that the caller’s roles include either RESEARCHER or STUDY_COORDINATOR. If they are a STUDY_COORDINATOR and not a RESEARCHER, the orgMembership is required.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -151,29 +144,56 @@ public void accept(JsonNode jsonNode) throws Exception {
...
+
+        // write the rest of the account summaries
+        while (!accountSummaries.isEmpty()) {
+            for (AccountSummary accountSummary : accountSummaries) {
+                String[] row = getAccountSummaryArray(accountSummary, csvHeaders);
+                csvFileWriter.writeNext(row);
+            }
+            // get next set of account summaries
+            offsetBy += PAGE_SIZE;
+            accountSummaries = bridgeHelper.getAccountSummariesForApp(appId, orgMembership, offsetBy, 0);
```
DwayneJengSage: Is this passing in 0 for page size? What would happen if we did this?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -101,7 +102,10 @@ public void accept(JsonNode jsonNode) throws Exception {
 
         String userId = request.getUserId();
         String appId = request.getAppId();
-        LOG.info("Received request for userId=" + userId + ", app=" + appId);
+        String password = request.getPassword();
+        String studyId = request.getStudyId();
```
DwayneJengSage, line 121: We need to verify that the caller's org sponsors the specified studyId. (Or the caller is a Researcher and is not associated with an org.)

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+
+            if (participantRoles == null || participantRoles.isEmpty() ||
+                    !(participantRoles.contains(Role.RESEARCHER) || participantRoles.contains(Role.STUDY_COORDINATOR))) {
+                LOG.info("User does not have a Researcher or Study Coordinator role.");
+                return;
+            }
+
+            if (participantRoles.contains(Role.STUDY_COORDINATOR) && !participantRoles.contains(Role.RESEARCHER)) {
+                orgMembership = participant.getOrgMembership();
+            }
```
DwayneJengSage: We should use the org for both Study Coordinator and Researcher. However, if Study Coordinator doesn't have an org, that's an error.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,169 @@
...
+        doReturn(participant).when(mockBridgeHelper).getParticipant(APP_ID, USER_ID, false);
+
+        when(mockBridgeHelper.getAccountSummariesForApp(anyString(), anyString(), anyInt(), anyInt(), anyString()))
+                .thenReturn(ImmutableList.of(makeAccountSummary(), makeAccountSummary(), makeAccountSummary()))
+                .thenReturn(new ArrayList<>());
+
+        doNothing().when(mockZipHelper).zipWithPassword(anyList(), any(File.class), anyString());
+
+        processor.accept(makeValidRequestNode());
+    }
```
DwayneJengSage: We should also verify the CSV written to the file.

We should also have a test case with just one page and a test case with multiple pages. Consider making PAGE_SIZE a paramter in the WorkerProcessor so that we can set it to a smaller page size to make the tests easier to run.

We should also verify calls to mockZipHelper and mockSesHelper.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private static AccountSummary makeAccountSummary() {
+        AccountSummary accountSummary = new AccountSummary();
+        accountSummary.putAttributesItem("firstName", "first-name");
+        accountSummary.putAttributesItem("lastName", "last-name");
+        accountSummary.putAttributesItem("email", "email@email.com");
+
+        return accountSummary;
+    }
```
DwayneJengSage: Consider taking an int or a string suffix as an input so that we can test this with rows that have different information.

Consider adding null values so that you can test that's handled correctly.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,169 @@
...
+        accountSummary.putAttributesItem("firstName", "first-name");
+
+        String[] headers = processor.getCsvHeaders(accountSummary);
+        assertEquals(headers.length, 1);
+        assertEquals("attributes", headers[0]);
+
+        String[] accountSummaryArray = processor.getAccountSummaryArray(accountSummary, headers);
+        assertEquals(accountSummaryArray.length, 1);
+        assertEquals("{\"firstName\":\"first-name\"}", accountSummaryArray[0]);
+    }
```
DwayneJengSage: Can we double check this? This test doesn't look right.

DwayneJengSage (changes requested):
Also need unit tests for SesHelper, BridgeHelper, and ZipHelper.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /** Given an Object, return a String array of all of its attribute keys
+     *  Visible for unit testing
+     */
+    String[] getCsvHeaders(Object accountSummary) {
+        JsonElement json = RestUtils.toJSON(accountSummary);
+        Set<String> keys = json.getAsJsonObject().keySet();
+        List<String> list = new ArrayList<>(keys);
+        return list.toArray(new String[0]);
+    }
```
DwayneJengSage: It looks like this might not work, because if an account doesn't have an attribute, it'll omit the field instead of providing that field with a null result. We may need to use Reflection instead to get all fields, or have a configurable list of fields

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /** Given an AccountSummary, return a String array of all of its attribute values
+     *  Visible for unit testing
+     */
+    String[] getAccountSummaryArray(AccountSummary accountSummary, String[] csvHeaders) {
+        JsonObject json = RestUtils.toJSON(accountSummary).getAsJsonObject();
+        ArrayList<String> list = new ArrayList<>();
+
+        for (int i = 0; i < csvHeaders.length; i++) {
+            String value = json.get(csvHeaders[i]) == null ? "" : json.get(csvHeaders[i]).toString();
```
DwayneJengSage: If the value is a string type, we should do .textValue() instead of .toString(), so we don't end up with a bunch of quotes everywhere.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
kllysng: Oh yes I see your point!! .textValue() doesn't come up, but .getAsString() does and it seems to work after testing a a couple of times. Please let me know if that's okay with you.

kllysng: I'm running into a problem after changing it to .getAsString(). When I call json.get(csvHeaders[i]), it will sometimes return a JsonElement, sometimes return a JsonPrimitive, and also JsonObject. The issue with when it returns a JsonObject is that .getAsString() doesn't work for it, and only .toString() does. I'll try to dig deeper..

DwayneJengSage: I got GSON and Jackson confused. For GSON, you'll want to check .isPrimitive() to be able to use .getAsString()

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+                return;
+            }
+
+            if (studyId != null && participantRoles.contains(Role.STUDY_COORDINATOR)) {
+
+                List<Study> studies = bridgeHelper.getStudiesForApp(appId, orgMembership, 0, pageSize);
+                if (studies.isEmpty()) {
+                    LOG.info("User's org does not sponsor the given studyId.");
+                    return;
+                }
```
DwayneJengSage: This doesn't do what you want it to do. You need to get all studies for the org and verify that the study list contains the specified studyId.

Also, for clarity, we should rename it getSponsoredStudiesForApp.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+                LOG.info("User does not have a Researcher or Study Coordinator role.");
+                return;
+            }
+
+            if (participantRoles.contains(Role.STUDY_COORDINATOR) && orgMembership == null) {
+                LOG.info("User is a Study Coordinator and does not have an org membership.");
+                return;
+            }
+
+            if (studyId != null && participantRoles.contains(Role.STUDY_COORDINATOR)) {
```
DwayneJengSage: If the caller is a Researcher and is part of an org, we need to verify the study as well.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -146,7 +175,35 @@ public void normalCase() throws Exception {
...
         processor.accept(makeValidRequestNode());
+
+        // verify that the csv is written to the file
+        verify(processor).writeAccountSummaries(any(), anyList(), anyInt(), anyString(), anyString(), anyString());
+
+        // verify that the file is zipped
+        verify(mockZipHelper).zipWithPassword(anyList(), any(File.class), anyString());
+
+        // verify that the email with attachment is sent
+        verify(mockSesHelper).sendEmailWithAttachmentToAccount(any(AppInfo.class), any(AccountInfo.class), anyString());
```
DwayneJengSage: Can we verify not just the methods that are called, but also the args?

Can we verify the files written and the CSV content?

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -107,37 +114,59 @@ public void emailNotVerified() throws Exception {
...
+        verifyNormalCase();
+    }
+
+    @Test
+    public void normalCaseMultiplePages() throws Exception {
+        // set page size
+        processor.setPageSize(5);
+
+        verifyNormalCase();
+    }
```
DwayneJengSage: This test case needs to mock multiple calls to Bridge and also verify that we page properly and write the result to the CSV properly.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -158,12 +215,36 @@ private static ObjectNode makeValidRequestNode() {
...
+        setVariableValueInObject(accountSummary, "firstName", "test-first-name");
+        setVariableValueInObject(accountSummary, "lastName", "test-last-name");
+        setVariableValueInObject(accountSummary, "email", "test@test.test");
+        setVariableValueInObject(accountSummary, "createdOn", now);
+        setVariableValueInObject(accountSummary, "status", null);
+        setVariableValueInObject(accountSummary, "appId", "test-app");
+        setVariableValueInObject(accountSummary, "orgMembership", "test-orgId");
 
         return accountSummary;
     }
```
DwayneJengSage: Consider taking an int or a string suffix as an input so that we can test this with rows that have different information.

DwayneJengSage: Are there any values that aren't primitives? Or are these always primitives? If the AccountSummary only contains primitives, we can simplify some of the code in DownloadParticipantRosterWorkerProcessor.

If AccountSummary can contain non-primitives, we should test that.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    String[] getAccountSummaryArray(AccountSummary accountSummary, String[] csvHeaders) {
+        JsonObject json = RestUtils.toJSON(accountSummary).getAsJsonObject();
+        ArrayList<String> list = new ArrayList<>();
+
+        for (int i = 0; i < csvHeaders.length; i++) {
+            JsonElement jsonValue = json.get(csvHeaders[i]);
+            String value = "";
+            if (jsonValue != null) {
+                value = (jsonValue.isJsonObject()) ? jsonValue.toString() : jsonValue.getAsString();
```
DwayneJengSage: What about JSON arrays?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -124,8 +131,18 @@ public void accept(JsonNode jsonNode) throws Exception {
                 return;
             }
 
-            if (participantRoles.contains(Role.STUDY_COORDINATOR) && !participantRoles.contains(Role.RESEARCHER)) {
-                orgMembership = participant.getOrgMembership();
+            if (participantRoles.contains(Role.STUDY_COORDINATOR) && orgMembership == null) {
+                LOG.info("User is a Study Coordinator and does not have an org membership.");
+                return;
+            }
```
DwayneJengSage, line 139: Unit test for Study Coordinator without org membership?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -124,8 +131,18 @@ public void accept(JsonNode jsonNode) throws Exception {
...
+
+            if (studyId != null && participantRoles.contains(Role.STUDY_COORDINATOR)) {
+
+                List<Study> studies = bridgeHelper.getStudiesForApp(appId, orgMembership, 0, pageSize);
+                if (studies.isEmpty()) {
+                    LOG.info("User's org does not sponsor the given studyId.");
+                    return;
+                }
```
DwayneJengSage: Unit test for study not sponsored by org?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -193,11 +210,14 @@ private void writeAccountSummaries(CSVWriter csvFileWriter, List<AccountSummary>
...
-        return list.toArray(new String[0]);
+    String[] getCsvHeaders() {
+        Field[] fields = AccountSummary.class.getDeclaredFields();
+        String[] headers = new String[fields.length];
+
+        for (int i = 0; i < fields.length; i++) {
+            headers[i] = fields[i].getName();
+        }
+        return headers;
     }
```
DwayneJengSage: We also want to exclude the following fields: appId, synapseUserId, orgMembership, type. Can probably create a static final Set<String> EXCLUDED_HEADERS and check the field against that set.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -229,7 +253,7 @@ private void zipFiles(File file, File zipFile, String password) throws IOExcepti
             zipHelper.zipWithPassword(ImmutableList.of(file), zipFile, password);
         } finally {
             zipStopwatch.stop();
-            LOG.info("Zipping to file " + zipFile.getAbsolutePath() + " took " +
+            LOG.info("Zipping to file " + zipFile + " took " +
```
DwayneJengSage: What happens if we log zipFile directly rather than zipFile.getAbsolutePath()?

src/main/java/org/sagebionetworks/bridge/udd/helper/ZipHelper.java
```diff
@@ -57,21 +63,48 @@ public void zip(List<File> fromList, File to) throws IOException {
...
-        zipParameters.setEncryptFiles(true);
-        zipParameters.setEncryptionMethod(EncryptionMethod.AES); // AES 256 by default
+    public void zipWithPassword(List<File> filesToAdd, File outputZipFile, String password) throws IOException {
+        ZipParameters zipParameters = buildZipParameters();
+        byte[] buffer = new byte[BUFFER_SIZE];
+        int readLen;
+
+        try (OutputStream bufferedOutputStream = new BufferedOutputStream(fileHelper.getOutputStream(outputZipFile))) {
+            net.lingala.zip4j.io.outputstream.ZipOutputStream zipOutputStream =
+                    new net.lingala.zip4j.io.outputstream.ZipOutputStream(bufferedOutputStream, password.toCharArray());
```
DwayneJengSage: We should put the ZipOutputStream inside the parentheses in the try. This automatically closes the Stream when the try block is done.

src/main/java/org/sagebionetworks/bridge/udd/helper/ZipHelper.java
```diff
@@ -57,21 +63,48 @@ public void zip(List<File> fromList, File to) throws IOException {
...
+            for (File file : filesToAdd) {
+                zipParameters.setFileNameInZip(file.getName());
+                zipOutputStream.putNextEntry(zipParameters);
+
+                try (InputStream inputStream = fileHelper.getInputStream(file)) {
+                    while ((readLen = inputStream.read(buffer)) != -1) {
+                        zipOutputStream.write(buffer, 0, readLen);
+                    }
```
DwayneJengSage: Consider using https://guava.dev/releases/19.0/api/docs/com/google/common/io/ByteStreams.html#copy(java.io.InputStream,%20java.io.OutputStream) so that you don't have to do your own stream buffer management.

src/main/java/org/sagebionetworks/bridge/udd/helper/ZipHelper.java
```diff
@@ -57,21 +63,48 @@ public void zip(List<File> fromList, File to) throws IOException {
...
+                }
+                zipOutputStream.closeEntry();
+            }
+        }
+    }
+
+    private net.lingala.zip4j.io.outputstream.ZipOutputStream initZip4jZipOutputStream(File outputZipFile, char[] password) throws IOException {
+        FileOutputStream fileOutputStream = new FileOutputStream(outputZipFile);
+        return new net.lingala.zip4j.io.outputstream.ZipOutputStream(fileOutputStream, password);
+    }
```
DwayneJengSage: Is this unused?

src/test/java/org/sagebionetworks/bridge/udd/helper/ZipHelperTest.java
```diff
@@ -90,4 +136,25 @@ public void testZip() throws Exception {
...
+                String fileName = localFileHeader.getFileName();
+                byte[] fileBytes = ByteStreams.toByteArray(zipInputStream);
+                String fileContent = new String(fileBytes, Charsets.UTF_8);
+                unzippedMap.put(fileName, fileContent);
+
+                localFileHeader = zipInputStream.getNextEntry();
+            }
+            return unzippedMap;
+        } catch (Exception e) {
+            return unzippedMap;
```
DwayneJengSage: If this throws, it probably indicates an error in the test somewhere. We should propagate the error back up and have the test fail.

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelperTest.java
```diff
@@ -624,6 +631,49 @@ public void redriveUpload_timeout() throws Exception {
...
+
+        when(mockWorkerApi.searchAccountSummariesForApp(eq(APP_ID), any(AccountSummarySearch.class))).thenReturn(mockCall);
+
+        when(mockClientManager.getClient(ForWorkersApi.class)).thenReturn(mockWorkerApi);
+
+        List<AccountSummary> retSummariesForApp = bridgeHelper.getAccountSummariesForApp(APP_ID, ORG_ID, 0,
+                BridgeHelper.MAX_PAGE_SIZE, STUDY_ID);
+
+
+        bridgeHelper.getAccountSummariesForApp(APP_ID, ORG_ID, 0, PAGE_SIZE, STUDY_ID);
```
DwayneJengSage: Minor: Extraneous call to getAccountSummariesForApp().

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelperTest.java
```diff
@@ -624,6 +631,49 @@ public void redriveUpload_timeout() throws Exception {
...
+    public void testGetAccountSummariesForApp() throws IOException {
+        // mock SDK search account summaries call
+        AccountSummaryList accountSummaryList = mock(AccountSummaryList.class);
+        when(accountSummaryList.getItems()).thenReturn(ImmutableList.of(new AccountSummary()));
+        Response<AccountSummaryList> response = Response.success(accountSummaryList);
+
+        Call<AccountSummaryList> mockCall = mock(Call.class);
+        when(mockCall.execute()).thenReturn(response);
+
+        when(mockWorkerApi.searchAccountSummariesForApp(eq(APP_ID), any(AccountSummarySearch.class))).thenReturn(mockCall);
```
DwayneJengSage: We should use a Captor to verify the params set in the AccountSummarySearch. We should verify the pageSize, offset, orgId, and studyId.

Also, we'll need 2 test cases, one for orgId and one for studyId.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelper.java
```diff
@@ -265,6 +267,30 @@ public Upload getUploadByRecordId(String recordId) throws IOException {
...
+
+        if (studyId != null) {
+            search.enrolledInStudyId(studyId);
+        } else if (orgId != null) {
+            search.orgMembership(orgId);
+        }
+
+        if (pageSize > 0) {
+            search.pageSize(PARTICIPANT_PAGE_SIZE);
+        }
```
DwayneJengSage: Bug: We set the file size to the default page size instead of the passed in page size.

src/test/java/org/sagebionetworks/bridge/udd/helper/SesHelperTest.java
```diff
@@ -89,6 +106,56 @@ public void testSendPresignedUrl() throws Exception {
...
+            assertEquals(rawMessage.getData().get(8), 'p');
+            assertEquals(rawMessage.getData().get(9), 'p');
+            assertEquals(rawMessage.getData().get(10), 'o');
+            assertEquals(rawMessage.getData().get(11), 'r');
+            assertEquals(rawMessage.getData().get(12), 't');
+            assertEquals(rawMessage.getData().get(13), '@');
+            assertEquals(rawMessage.getData().get(14), 's');
+            assertEquals(rawMessage.getData().get(15), 'a');
+            assertEquals(rawMessage.getData().get(16), 'g');
+            assertEquals(rawMessage.getData().get(17), 'e');
```
DwayneJengSage: You can probably do new String(rawMessage.getData().array()) to convert this into a String.

src/test/java/org/sagebionetworks/bridge/udd/helper/SesHelperTest.java
```diff
@@ -89,6 +106,56 @@ public void testSendPresignedUrl() throws Exception {
...
+
+            // ...
```
DwayneJengSage: Looks like the rest of this test needs to be finished.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
kllysng: Aren't JSON arrays able to call .getAsString()? Or am I misunderstanding your question?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
kllysng: There doesn't appear to be a difference: 
Screen Shot 2021-04-13 at 3 26 13 PM

src/test/java/org/sagebionetworks/bridge/udd/helper/ZipHelperTest.java
kllysng: The reason I did this was because LocalFileHeader (zip4j's version of ZipEntry) frustratingly does not return null when there is no more file to read, instead it will throw an error. I spent a few hours grappling with this to no avail and I haven't been able to find a way to deal with this. The other way of unzipping a password protected zip file using this library is done in one line but it extracts it to a directory only. Does this make sense?

src/test/java/org/sagebionetworks/bridge/udd/helper/SesHelperTest.java
kllysng: For some reason .toString() converts into "java.nio.HeapByteBuffer[pos=0 lim=1196 cap=1196]", but I found a different way of doing this.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+
+            if (participant.getEmail() == null || !participant.isEmailVerified()) {
+                LOG.info("User does not have a validated email address.");
+                return;
+            }
+
+            int offsetBy = 0;
+            // get first page of account summaries
+            List<AccountSummary> accountSummaries = bridgeHelper.getAccountSummariesForApp(appId, orgMembership,
+                    offsetBy, pageSize, studyId);
```
DwayneJengSage: Also, Alx mentioned that the Account Summary doesn't contain all fields, so we need to call getParticipantByIdForApp() to get all the Account fields.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,371 @@
...
+
+        // process
+        processor.process(request, csvFile, zipFile);
+
+        // verify file contents
+        String csvContent = new String(fileHelper.getBytes(csvFile), StandardCharsets.UTF_8);
+        assertEquals(EXPECTED_CSV_CONTENT, csvContent);
+
+        // verify that the csv is written to the file
+        verify(processor).writeStudyParticipants(any(CSVWriter.class), eq(studyParticipants), eq(0), eq(APP_ID), eq(ORG_ID), anyString());
```
DwayneJengSage: Can we verify the studyId being passed in? (Which if there is no studyId, it should be null.)

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,371 @@
...
+
+        // verify that the file is zipped
+        verify(mockZipHelper).zipWithPassword(anyList(), any(File.class), anyString());
+
+        // verify that the email with attachment is sent
+        verify(mockSesHelper).sendEmailWithAttachmentToAccount(any(AppInfo.class), any(AccountInfo.class), anyString());
```
DwayneJengSage: Can we verify the args to these two calls?

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,371 @@
...
+                } else if (count == 1) {
+                    count++;
+                    return ImmutableList.of(participant2);
+                } else if (count == 2) {
+                    count++;
+                    return ImmutableList.of(participant3);
+                }
+                return ImmutableList.of();
+            }
+        }).when(mockBridgeHelper).getStudyParticipantsForApp(eq(APP_ID), eq(ORG_ID), anyInt(), eq(1), anyString());
```
DwayneJengSage: Since we control the offset, it might be easier to have 3 mocks to getStudyParticipantsForApp() with offset 0, offset 1, and offset 2 instead of using a doAnswer().

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -0,0 +1,371 @@
...
+        assertEquals("", studyParticipantArrayArray[16]);
+        assertEquals("test@test.test", studyParticipantArrayArray[17]);
+        assertEquals("", studyParticipantArrayArray[18]);
+        assertEquals("", studyParticipantArrayArray[19]);
+        assertEquals("", studyParticipantArrayArray[20]);
+        assertEquals("", studyParticipantArrayArray[21]);
+    }
+
+    @Test
+    public void process() throws Exception {
```
DwayneJengSage: Should we also have a test case with studyId?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -211,27 +213,38 @@ void writeAccountSummaries(CSVWriter csvFileWriter, List<AccountSummary> account
...
-                value = (jsonValue.isJsonObject()) ? jsonValue.toString() : jsonValue.getAsString();
+                if (jsonValue.isJsonPrimitive()) {
+                    value = jsonValue.getAsString();
+                } else if (jsonValue.isJsonObject()) {
+                    value = jsonValue.toString();
+                } else if (jsonValue.isJsonArray()) {
+                    for (JsonElement element : (JsonArray) jsonValue) {
+                        value += element.getAsString();
+                    }
+                }
```
DwayneJengSage: Would it be better to use .toString() for JSON arrays? Also, do we have a test case for this?

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -179,28 +183,26 @@ public void accept(JsonNode jsonNode) throws Exception {
                 throw new RuntimeException(ex);
             }
         } finally {
-            cleanupFiles(csvFile, zipFile, tmpDir);
-            LOG.info("request took " + requestStopwatch.elapsed(TimeUnit.SECONDS) +
-                    " seconds for userId =" + userId + ", app=" + appId);
+            //cleanupFiles(csvFile, zipFile, tmpDir);
```
DwayneJengSage: Un-comment cleanupFiles()

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelper.java
```diff
@@ -279,18 +280,32 @@ public Upload getUploadByRecordId(String recordId) throws IOException {
...
+    public List<Study> getSponsoredStudiesForApp(String appId, String orgId, int offsetBy, int pageSize) throws IOException {
         return clientManager.getClient(ForWorkersApi.class).getSponsoredStudiesForApp(appId, orgId, offsetBy, pageSize)
                 .execute().body().getItems();
     }
 
+    private List<StudyParticipant> getStudyParticipantsFromAccountSummaries(List<AccountSummary> accountSummaries) throws IOException {
+        List<StudyParticipant> participants = new ArrayList<>();
+        for (AccountSummary accountSummary : accountSummaries) {
+            StudyParticipant participant = clientManager.getClient(ForWorkersApi.class)
+                    .getParticipantByIdForApp(accountSummary.getAppId(), accountSummary.getId(), false).execute().body();
```
DwayneJengSage: I think we want to include consent histories as well.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelper.java
```diff
@@ -279,18 +280,32 @@ public Upload getUploadByRecordId(String recordId) throws IOException {
...
+            participants.add(participant);
+        }
```
DwayneJengSage: Add a short sleep (maybe 20ms) so that we don't burn out the server.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,287 @@
...
+                String[] row = getStudyParticipantArray(studyParticipant, csvHeaders);
+                csvFileWriter.writeNext(row);
+            }
+            // get next set of study participants
+            offsetBy += pageSize;
+            studyParticipants = bridgeHelper.getStudyParticipantsForApp(appId, orgMembership, offsetBy, pageSize, studyId);
+
+            // avoid burning out Bridge Server
+            Thread.sleep(THREAD_SLEEP_INTERVAL);
+        }
```
DwayneJengSage: Since there could be lots of participants (thousands or tens of thousands), consider adding a log message every page so that we can look at the logs and verify whether the worker is making progress.

src/main/java/org/sagebionetworks/bridge/udd/helper/SesHelper.java
```diff
@@ -123,4 +179,63 @@ private void sendEmailToAccount(AppInfo appInfo, AccountInfo accountInfo, Body b
...
+        message.setSubject(subjectStr, CHARSET_UTF_8);
+        message.setFrom(new InternetAddress(fromAddress));
+        message.setRecipient(javax.mail.Message.RecipientType.TO, new InternetAddress(destination));
+
+        // create multipart/alternative child container, wrapper for the HTML and text parts
+        MimeMultipart messageBody = new MimeMultipart(CONTENT_SUBTYPE_ALTERNATIVE);
+        MimeBodyPart wrap = new MimeBodyPart();
+
+        MimeBodyPart textPart = new MimeBodyPart();
+        textPart.setContent(ATTACHMENT_BODY_TEMPLATE_TEXT, CONTENT_TYPE_TEXT_HTML);
```
DwayneJengSage: This should probably be Mime Type text/plain.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
kllysng: @DwayneJengSage Do you know if there's a test studyId that I can provide that will be considered sponsored by the given test org id? Otherwise, it will fail because of the sponsorship requirement.

DwayneJengSage: It comes from getSponsoredStudies().

kllysng: Makes so much sense. Thank you.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
kllysng: @DwayneJengSage I guess StudyParticipant doesn't have any array attributes, so I don't think that the JsonArray branch will be reached. I think I should remove it. Is that okay with you?

DwayneJengSage: Sounds reasonable.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -0,0 +1,293 @@
...
+                throw new RuntimeException(ex);
+            }
+        } finally {
+            cleanupFiles(csvFile, zipFile, tmpDir);
+        }
+    }
+
+    /** Write all of the study participants to the CSV file */
+    void writeStudyParticipants(CSVWriter csvFileWriter, List<StudyParticipant> studyParticipants, int offsetBy,
+                                String appId, String orgMembership, String studyId, int numParticipants) throws IOException, InterruptedException {
```
DwayneJengSage: What is numParticipants used for, and why does it need to be passed in as an input to this method?

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
```diff
@@ -192,38 +205,75 @@ public void process() throws Exception {
...
+        processor.setPageSize(1); // the processor will need to make 3 Bridge API calls
+        processor.process(request, csvFile, zipFile, tmpDir);
 
         // verify file contents
         String csvContent = new String(fileHelper.getBytes(csvFile), StandardCharsets.UTF_8);
         assertEquals(EXPECTED_CSV_CONTENT, csvContent);
 
         // verify that the csv is written to the file
-        verify(processor).writeStudyParticipants(any(CSVWriter.class), eq(studyParticipants), eq(0), eq(APP_ID), eq(ORG_ID), anyString());
+        verify(processor).writeStudyParticipants(any(CSVWriter.class), anyList(), eq(0), eq(APP_ID), eq(ORG_ID), eq(null), anyInt());
```
DwayneJengSage: I think there's some confusion here as to why we're calling verify() on the method that we're testing.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -234,19 +244,15 @@ void writeStudyParticipants(CSVWriter csvFileWriter, List<StudyParticipant> stud
 
         for (int i = 0; i < csvHeaders.length; i++) {
             JsonElement jsonValue = json.get(csvHeaders[i]);
-            String value = "";
+            StringBuilder value = new StringBuilder();
```
DwayneJengSage: I don't think we need to use a StringBuilder if we're just adding one string and then immediately calling .toString() on it.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
kllysng: I confused myself when I was doing this but I was responding to the comment about adding a log message every page. I've removed the numParticipants logic and added a simple log message.

src/test/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessorTest.java
kllysng: I think I understand what you're saying - I just replaced anyList() with an ImmutableList.of(participant1), which constitutes the first "page" of study participants to be written. Now the params are being checked.

DwayneJengSage (approved):

PR #107, opened 2021-08-09
Exporter 3.0

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-3069

See https://github.com/Sage-Bionetworks/BridgeServer2/pull/390

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,11 @@
             <version>0.22.1</version>
         </dependency>
         <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.15</version>
+        </dependency>
+        <dependency>
             <groupId>joda-time</groupId>
             <artifactId>joda-time</artifactId>
             <version>2.9.4</version>
@@ -98,7 +103,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>bridge-base</artifactId>
-            <version>2.7.26</version>
+            <version>2.7.30</version>
         </dependency>
         <dependency>
             <groupId>org.sagebionetworks</groupId>
@@ -108,7 +113,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.22.1</version>
+            <version>0.22.18</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>bridge-base</artifactId>
-            <version>2.7.30</version>
+            <version>2.7.31</version>
         </dependency>
         <dependency>
             <groupId>org.sagebionetworks</groupId>
@@ -113,7 +113,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.22.18</version>
+            <version>0.22.24</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
@@ -152,7 +152,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>BridgeTestUtils</artifactId>
-            <version>1.0</version>
+            <version>1.5</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
@@ -12,6 +12,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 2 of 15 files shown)


src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+            metadata.addUserMetadata(metadataEntry.getKey(), metadataEntry.getValue());
+        }
+
+        return metadata;
+    }
+
+    private void exportToSynapse(String appId, Exporter3Configuration exporter3Config, Upload upload,
+            ExportContext exportContext)
+            throws SynapseException {
+        // Exports are folderized by calendar date (YYYY-MM-DD). Create that folder if it doesn't already exist.
```
DwayneJengSage, line 335: Link Jira to Synapse folder limits so we document what our limits are.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+        fileEntity = synapseHelper.createEntityWithRetry(fileEntity);
+        String fileEntityId = fileEntity.getId();
+
+        // Add annotations.
+        Map<String, AnnotationsValue> annotationMap = new HashMap<>();
+        for (Map.Entry<String, String> metadataEntry : exportContext.getMetadataMap().entrySet()) {
+            AnnotationsValue value = new AnnotationsValue();
+            value.setType(AnnotationsValueType.STRING);
+            value.setValue(ImmutableList.of(metadataEntry.getValue()));
+            annotationMap.put(metadataEntry.getKey(), value);
```
DwayneJengSage, line 374: Are there types that researchers will want to query as an int? If so, do we need to store type information, and how?

pom.xml
```diff
@@ -152,7 +152,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>BridgeTestUtils</artifactId>
-            <version>1.0</version>
+            <version>1.5</version>
```
DwayneJengSage, line 155: Double check: Do I have a pending change in TestUtils.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -87,7 +89,7 @@ public void setMetadataMap(Map<String, String> metadataMap) {
     static final String METADATA_KEY_HEALTH_CODE = "healthCode";
     static final String METADATA_KEY_RECORD_ID = "recordId";
     static final String METADATA_KEY_UPLOADED_ON = "uploadedOn";
-
+    private static final Pattern METADATA_NAME_REPLACEMENT_PATTERN = Pattern.compile("[^\\w\\.]");
```
DwayneJengSage, line 94: Comment here, alphanumeric, periods, and underscores.

DwayneJengSage, line 94: Should we namespace our annotations (for example with a "Bridge." prefix).

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -152,26 +156,40 @@ public void accept(JsonNode jsonNode) throws IOException, PollSqsWorkerBadReques
...
             return;
         }
 
         // Set the exportedOn time on the record. This will be propagated to both S3 and Synapse.
-        HealthDataRecordEx3 record = bridgeHelper.getHealthDataRecordForExporter3(recordId);
+        HealthDataRecordEx3 record = bridgeHelper.getHealthDataRecordForExporter3(appId, recordId);
         record.setExportedOn(DateUtils.getCurrentDateTime());
 
         // Copy the file to the raw health data bucket. This includes folderization.
-        Upload upload = bridgeHelper.getUploadByRecordId(recordId);
```
DwayneJengSage: File Jira for getUploadByRecordId doesn't work for Exporter 3.0.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -345,18 +374,18 @@ private void exportToSynapse(String appId, Exporter3Configuration exporter3Confi
...
     }
 
     private String getFilenameForUpload(Upload upload) {
         return upload.getUploadId() + '-' + upload.getFilename();
     }
 
-    private String getCalendarDateForUpload(Upload upload) {
-        LocalDate localDate = upload.getCompletedOn().withZone(Constants.LOCAL_TIME_ZONE).toLocalDate();
+    private String getCalendarDateForRecord(HealthDataRecordEx3 record) {
+        LocalDate localDate = record.getExportedOn().withZone(Constants.LOCAL_TIME_ZONE).toLocalDate();
```
DwayneJengSage: Revert this change. The upload should be bucketed based on when the app sent it to Bridge (completedOn), not when it was Exported to Synapse (exportedOn), because it could be exported multiple times or at a later date (eg a export redrive).

src/main/java/org/sagebionetworks/bridge/workerPlatform/config/SpringConfig.java
```diff
@@ -69,6 +68,7 @@
 // See http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/credentials.html and
 // http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-setup.html#set-up-creds for more info.
 @ComponentScan({
+        "org.sagebionetworks.bridge.exporter3",
```
DwayneJengSage: Can we update this to just ComponentScan("org.sagebionetworks.bridge") so we don't have to update this every time?

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
DwayneJengSage: Filed https://sagebionetworks.jira.com/browse/BRIDGE-3079

PR #116, opened 2021-12-22
BRIDGE-2606 Export Participant Versions

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2606

See also Sage-Bionetworks/BridgeServer2#501

Still need to write unit tests for BridgeHelper and Ex3ParticipantVersionWorkerProcessor


src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -313,8 +309,14 @@ private ExportContext copyUploadToHealthDataBucket(String appId, Upload upload,
         Map<String, String> recordMetadataMap = record.getMetadata();
         if (recordMetadataMap != null) {
             for (Map.Entry<String, String> metadataEntry : record.getMetadata().entrySet()) {
+                // Replace invalid characters in metadata name.
```
DwayneJengSage, line 305: Move this to before Bridge-Specific Metadata to make sure Bridge wins out if there is a conflict.

alxdarksage (approved):

PR #119, opened 2022-01-19
Update participant roster

alxdarksage:
Some issues addressed:

- Check of study access was only pulling one page of studies...in MTB at this point it's possible the study could be some pages into the list, so this is fixed with the introduction of a PagedResourceList iterator that keeps paging until all records are visited;
- Added some roles that I think should be able to generate the roster (admin and superadmin);
- Most importantly, the zip file Is uploaded to an S3 bucket created by BridgeServer2, and an access link is generated and emailed to the original requestor of the data. The zip file is produced on disk and I think uploaded from disk, so this should reduce memory consumption, and it will avoid SES rejections of email messages that are too big.

There are no integration tests for the participant roster worker processor...I wrote one to manually test but haven't checked it in because all it can do is verify a successful call to BridgeServer2 to kick off the process.


src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
```diff
@@ -122,60 +165,41 @@ void process(DownloadParticipantRosterRequest request, File csvFile, File zipFil
...
 
             zipFiles(csvFile, zipFile, password);
-            sesHelper.sendEmailWithAttachmentToAccount(appInfo, accountInfo, zipFile.getAbsolutePath());
+            
+            ObjectMetadata metadata = new ObjectMetadata();
+            metadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
+            metadata.setContentType(CONTENT_TYPE_ZIP);
+            metadata.setContentDisposition("attachment; filename=\""+ZIP_FILE_NAME+"\"");
+            
+            String s3Key = appId + "/" + studyId + "/" + getNextToken() + getNextToken() + "/" + ZIP_FILE_NAME;
```
alxdarksage: can you call getNextToken which is just the right size without doing it twice?

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The iterator fails on the first page that encounters an error, ending the iteration. 
+     */
+    private void getNextPage() {
+        try {
+            page = func.apply(offsetBy, pageSize);
+        } catch (IOException e) {
+            page = ImmutableList.of();
+            LOG.error("Error while calling paged iterator", e);
+        }
```
alxdarksage, line 84: Bridge exception - if 400-499 stop, if it's 500, retry a limited number of times.

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelperTest.java
```diff
@@ -679,36 +679,7 @@ public void redriveUpload_timeout() throws Exception {
     }
 
     @Test
-    public void testGetStudyParticipantsForAppStudyId() throws IOException, IllegalAccessException, InterruptedException {
```
alxdarksage, line 698: Double-check you didn't accidentally delete something.

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIteratorTest.java
```diff
@@ -0,0 +1,131 @@
...
+        
+        Call<StudyList> call1 = mock(Call.class);
+        when(call1.execute()).thenReturn(Response.success(page1));
+        
+        when(mockWorkersApi.getSponsoredStudiesForApp("api", "orgId", start, 10)).thenReturn(call1);
+    }
+    
+    @Test
+    public void test() throws Exception {
+        List<String> expectedValues = new ArrayList<>();
```
alxdarksage: expectedIdentifiers, returnedIdentifiers

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: BridgeSDKException is a RuntimeException, which you're not even catching.

src/main/java/org/sagebionetworks/bridge/participantroster/DownloadParticipantRosterWorkerProcessor.java
alxdarksage: Removed extra call to getNextToken()

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: Updated with a specific list of HTTP error codes that are potentially recoverable, and on these error codes, iterator retries with exponential back off. Otherwise, it fails with logging and indicates to the consumer that there are no more elements in the iteration.

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/BridgeHelperTest.java
alxdarksage: Yes, this diff is confusing because I renamed a test to reflect the method it was calling on the worker (testGetSponsoredStudiesForApp) AND removed a utility method since there is only one test of that method now, for a given studyId (and not orgId). Test coverage is still all green.

src/test/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIteratorTest.java
alxdarksage: renamed

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
```diff
@@ -1,47 +1,106 @@
...
+     * Timeout (408), Too Many Requests (429), Bad Gateway (502), Service
+     * Unavailable (503), or Gateway Timeout (504), it will retry up to three times
+     * with exponential back-off. Other exceptions cause the iterator to stop iteration.
+     */
+    private void loadNextPage() {
+        page = ImmutableList.of(); // a failure will stop iteration.
+        for (int i=0; i <= NUMBER_RETRIES; i++) {
+            if (i > 0) { // this is a retry
+                try {
+                    long timeout = (long)Math.pow(retryDelayInSeconds, i);
```
DwayneJengSage: With the default retry delay of 5, this means that the first retry delay will be 5 seconds, then 25 seconds, then 125 seconds. This by itself isn't a problem, but it's essentially super-exponential backoff.

A more traditional approach would be 5 seconds, then 10 seconds, then 20 seconds, or retryDelayInSeconds*Math.pow(2, i);

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
```diff
@@ -1,47 +1,106 @@
...
+                    LOG.error("Non-recoverable exception while calling paged iterator", e);
+                    break;
+                } else if (i == NUMBER_RETRIES) {
+                    LOG.info("Aborting request after "+NUMBER_RETRIES+" retries", e);
+                    break;
+                }
+            } catch(Throwable t) {
+                LOG.error("Non-recoverable exception while calling paged iterator", t);
+                break;
+            }
```
DwayneJengSage, line 83: Should we retry on IOException? I can imagine that a network error might cause an IOException, and that we would want to retry on this.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
```diff
@@ -1,47 +1,106 @@
...
+                page = func.apply(offsetBy, pageSize);
+                break;
+            } catch(BridgeSDKException e) {
+                if (nonRecoverableException( e.getStatusCode() )) {
+                    LOG.error("Non-recoverable exception while calling paged iterator", e);
+                    break;
+                } else if (i == NUMBER_RETRIES) {
+                    LOG.info("Aborting request after "+NUMBER_RETRIES+" retries", e);
+                    break;
+                }
```
DwayneJengSage: Is this block necessary? If i == NUMBER_RETRIES, the for loop would simply end after this iteration. By moving the LOG.info statement to after the for loop, we can apply the retry logic exceptions other than BridgeSDKException without having to duplicate this if statement across different catch blocks.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
```diff
@@ -1,47 +1,106 @@
...
+                break;
+            }
+        }
         this.offsetBy += pageSize;
         this.nextIndex = 0;
     }
+    
+    private boolean nonRecoverableException(int statusCode) {
+        return statusCode != 408 && statusCode != 429 && statusCode != 502 && statusCode != 503 && statusCode != 504;
+    }
```
DwayneJengSage: Should we add 500 to this list? HTTP 500 Internal Server Error would likely mean a bug or a transient error in Bridge Server, which may be recoverable. Worst case scenario, it would try 3 times before giving up.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: Changed this to back off less aggressively.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: I removed it, we don't need to log this. I was logging it to manually test but we'll know that the iterator stops retrying after N retries.

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: I can add 500, but I wouldn't normally expect it to be recoverable...

src/main/java/org/sagebionetworks/bridge/workerPlatform/bridge/PagedResourceIterator.java
alxdarksage: Hmm, yes. I'll add this.

alxdarksage:
Also updated the comments.

DwayneJengSage (approved):

PR #120, opened 2022-01-21
Add worker to cache weekly adherence reports for users.

alxdarksage:
This merged updates to the participant roster worker; after that is merged, this PR will be clearer.


src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportRequest.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+    public String getStudyId() {
+        return studyId;
+    }
+    public void setStudyId(String studyId) {
+        this.studyId = studyId;
+    }
+    
+    
+
```
alxdarksage: Remove whitespace

alxdarksage: This could be a Map<String, List<String>> apps -> studies in those apps

DwayneJengSage: Maybe a map of String to Set of string. A set instead of a list for studyIDs means we can use a contains() in constant(ish) time instead of linear time.

src/test/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessorTest.java
```diff
@@ -0,0 +1,219 @@
...
+        Call<WeeklyAdherenceReport> failedReportCall = mockCall(failedReport);
+        
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target1", "user1")).thenReturn(successfulReportCall);
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target2", "user1")).thenReturn(successfulReportCall);
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target3", "user1")).thenReturn(successfulReportCall);
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target1", "user2")).thenReturn(successfulReportCall);
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target2", "user2")).thenReturn(failedReportCall);
+        when(mockWorkersApi.getWeeklyAdherenceReportForWorker("target", "study-target3", "user2")).thenReturn(failedReportCall);
+        
+        processor.accept(null); // we're doing everything, and that's it.
```
DwayneJengSage: Does this actually work? If we ask Jackson to parse a null, will it give us the default POJO with no attributes?

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+            LOG.info("Weekly adherence report caching took " + requestStopwatch.elapsed(TimeUnit.SECONDS) + " seconds");
+        }
+    }
+    
+    private void process(WeeklyAdherenceReportRequest request) throws Exception {
+        ForWorkersApi workersApi = clientManager.getClient(ForWorkersApi.class);
+        
+        List<App> appList = null;
+        if (request.getAppId() != null) {
+            appList = ImmutableList.of( bridgeHelper.getApp(request.getAppId()) );
```
DwayneJengSage: We should add a small unit test to verify that we filter appIDs if an appID is specified.

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+                AccountSummarySearch search = new AccountSummarySearch().offsetBy(ob).pageSize(ps).enrollment(ENROLLED);
+                return workersApi.searchAccountSummariesForApp(app.getIdentifier(), search).execute().body().getItems();
+            }, PAGE_SIZE);
+            
+            while (acctIterator.hasNext()) {
+                AccountSummary summary = acctIterator.next();
+                for (String studyId : summary.getStudyIds()) {
+                    if (request.getStudyId() != null && !studyId.equals(request.getStudyId())) {
+                        continue;
+                    }
```
DwayneJengSage: We should add a short unit test to test that this filters out studies correctly.

DwayneJengSage (changes requested):
Changes look good. Just a few comments about unit tests for branch coverage for filter cases.

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportRequest.java
alxdarksage: Yes. This has been updated, and I've added a test that the processor correctly skips over apps and studies that are not in the map (when it is not empty).

src/test/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessorTest.java
alxdarksage: Not since I added a request object...tests updated to work and testing the payload is utilized.

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
alxdarksage: Added verification that we don't call for an app when not specified.

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
alxdarksage: Added

DwayneJengSage (approved):

PR #138, opened 2022-05-02
Execute adherence multiple times per day determined by the local time zone for each study (BRIDGE-3214)

alxdarksage:


src/test/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessorTest.java
```diff
@@ -305,6 +319,163 @@ public void limitsScanToProvidedAppsAndStudies() throws Exception {
         verify(mockWorkersApi, never()).getWeeklyAdherenceReportForWorker("app1", "study1b", "user1");
         verify(mockWorkersApi, never()).getWeeklyAdherenceReportForWorker("app2", "study2b", "user1");
     }
+    
+    @SuppressWarnings("unchecked")
+    @Test
+    public void onlyRunsOnTargetedHourOfDay() throws Exception {
```
DwayneJengSage: It looks like this test includes a lot of things that are already covered by the previous test. Can we simplify this test?

DwayneJengSage: Also, all the test cases get their time zone from the study. Should we have a test case that gets the time zone from the request?

DwayneJengSage (approved):
Overall looks good. A few questions about the unit test.

src/test/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessorTest.java
alxdarksage: I've tried to create tests that are just demonstrating the use of the time zones and reporting hours, and taken out code that verifies the request of apps and users. Also I've renamed those tests and added more explicit comments about what they're testing. Hopefully it's clearer.

alxdarksage: Added this case

DwayneJengSage (approved):

PR #150, opened 2022-08-23
truncate invalid long languages in participant versions

DwayneJengSage:
Very minor bug fix here: Apparently, some people are submitting invalid languages (eg "Spanish" instead of "es-mx"). This causes the participant version to fail to export entirely.

The simple fix is to just truncate the language.


jnelsonjava (approved):
Looks good.

PR #151, opened 2022-08-29
Add error handling to WeeklyAdherenceReportWorkerProcessor (BRIDGE-3332)

jnelsonjava:
See [BRIDGE-3332](https://sagebionetworks.jira.com/browse/BRIDGE-3332)

Wrapping weekly adherence report requests in try/catch block to prevent all reports from failing due to individual account issues.


imranmaj (approved):

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
```diff
@@ -150,8 +151,16 @@ private void process(WeeklyAdherenceReportRequest request) throws Exception {
...
-                    WeeklyAdherenceReport report = workersApi.getWeeklyAdherenceReportForWorker(
-                            app.getIdentifier(), studyId, summary.getId()).execute().body();
+                    WeeklyAdherenceReport report = null;
+                    try {
+                        report = workersApi.getWeeklyAdherenceReportForWorker(
+                                app.getIdentifier(), studyId, summary.getId()).execute().body();
+                    } catch (BridgeSDKException e) {
+                        // Catching this specific error to prevent a single account issue from
+                        // stopping the remaining reports
+                        LOG.error("Error while processing adherence job", e);
```
DwayneJengSage: Minor: I would include app, study, and user ID in the error message. This will save us a step when looking at the logs.

DwayneJengSage (approved):
Looks good. One minor comment about the error message.

src/main/java/org/sagebionetworks/bridge/adherence/WeeklyAdherenceReportWorkerProcessor.java
jnelsonjava: That's a good call! I'll add that before merging.

PR #154, opened 2022-09-27
Add worker support for exporting demographics with participant versions

imranmaj:
https://sagebionetworks.jira.com/browse/BRIDGE-3338

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/exporter3/Ex3ParticipantVersionWorkerProcessorTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/exporter3/Ex3ParticipantVersionWorkerProcessorTest.java
@@ -7,6 +7,7 @@ import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.isNull;
 import static org.mockito.Matchers.same;
 import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.never;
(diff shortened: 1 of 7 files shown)


src/test/java/org/sagebionetworks/bridge/exporter3/Ex3ParticipantVersionWorkerProcessorTest.java
```diff
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.exporter3;
 
+import static org.junit.Assert.assertNotNull;
```
DwayneJengSage: Minor: We don't want to mix JUnit with TestNG. It matters for annotations, but not for asserts, but regardless, we'd like to be consistent and stick to TestNG.

(Note: BridgeIntegrationTests is the exception, since that was originally written in JUnit and has not yet been migrated.)

DwayneJengSage (approved):
Looks good. Just one comment about imports.

src/test/java/org/sagebionetworks/bridge/exporter3/Ex3ParticipantVersionWorkerProcessorTest.java
imranmaj: Ah, that was an accident, thanks

PR #170, opened 2023-02-15
Move S3EventNotificationCallback from Exporter to Worker

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2225

This is mostly copied from Exporter 2.0. Two major differences

1. I was wrong about the Antivirus Trigger SNS topic having a different format. Apparently, I just neglected to click a checkbox when I manually set up the subscription. This change reverts the format change.
2. The Antivirus Trigger SNS topic listens to multiple buckets, not just the upload bucket. There is code in here to filter only to the upload bucket.


jnelsonjava (approved):
Looks good.

PR #176, opened 2023-09-21
Fix for DHP-1019: Add appID to upload metadata

nategbrown9:


DwayneJengSage (approved):
Looks good.

Make sure you run `mvn verify` in your local environment. This will run all the checks, including unit tests, FindBugs (static code analysis), and Jacoco Unit Test coverage. None of these should fail given how simple the changes are, but it's good engineering practice.

Also, I think you need to update the integ tests as well https://github.com/Sage-Bionetworks/BridgeWorkerIntegrationTests/blob/develop/src/test/java/org/sagebionetworks/bridge/exporter/integration/Exporter3Test.java#L1269

Integ tests should be pretty straightforward to set up. The only thing that's missing from the documentation is that you need to set up an admin account for Integ Tests separate from the worker account that the Worker uses.

PR #178, opened 2023-11-02
DHP-1072: Add JSON to table row to exporter3

nategbrown9:


src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
     }
 
+    UploadTableRow getUploadTableRow(Upload upload, HealthDataRecordEx3 record, String recordId, Map<String, String> metadataMap )
+            throws IOException {
+        UploadTableRow tableRow = null;
+        String instanceGuid = metadataMap.get(METADATA_KEY_INSTANCE_GUID);
+        String assessmentGuid = metadataMap.get("assessmentGuid");
+        String appID = record.getAppId();
+
+        if (instanceGuid != null) {
```
DwayneJengSage, line 564: We check instanceGuid here, but we never use it. Do you mean assessmentGuid?

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+            if (!Strings.isNullOrEmpty(startedOn)) {
+                createdOn = DateTime.parse(metadataMap.get("startedOn"));
+            }
+            tableRow.setCreatedOn(createdOn);
+            tableRow.setRecordId(recordId);
+            tableRow.setHealthCode(record.getHealthCode());
+            tableRow.setParticipantVersion(record.getParticipantVersion());
+            Map<String, String> metaData = new HashMap<>();
+            metaData.put("clientInfo", record.getClientInfo());
+            String[] metadataKeys = new String[] {
```
DwayneJengSage, line 578: We can make metadataKeys a private static final in the class.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+                    metaData.put(key, value);
+                }
+            }
+            tableRow.setMetadata(metaData);
+
+            boolean isTestData = false;
+            HealthDataRecord healthData = upload.getHealthData();
+            if (healthData != null) {
+                isTestData = healthData.getUserDataGroups().contains("test_user");
+                metaData.put("externalId", healthData.getUserExternalId());
```
DwayneJengSage, line 594: We had a request to remove the study ID suffix from the external ID.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+                metaData.put("dataGroups", String.join(",", healthData.getUserDataGroups()));
+            }
+            tableRow.setTestData(isTestData);
+
+            Assessment assessment = bridgeHelper.getAssessment(appID, assessmentGuid);
+            AssessmentConfig assessmentConfig = bridgeHelper.getAssessmentConfig(appID, assessmentGuid);
+            AssessmentSummarizerProvider summarizerProvider = new AssessmentSummarizerProvider();
```
DwayneJengSage, line 601: We can make this a Spring Component and inject it using Spring.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+            AssessmentSummarizer summarizer = summarizerProvider.getSummarizer(assessment, assessmentConfig);
+            if (summarizer != null) {
+                File tempDir = null;
+                try {
+                    tempDir = fileHelper.createTempDir();
+                    File downloadedFile = fileHelper.newFile(tempDir, upload.getUploadId());
+                    s3Helper.downloadS3File(rawHealthDataBucket, upload.getUploadId(), downloadedFile);
```
DwayneJengSage, line 608: The filename won't just be upload ID. You'll need to call getRawS3KeyForUpload() to get the S3 Key.

There's also a weird edge case where EX3 isn't enabled for the app but is inabled for the study, in which case you won't have the information you need to download the file. This won't affect Open Bridge, but could affect other apps in the future. We should file a Jira to track this even if we choose not to handle it for 2023.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +554,89 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+                        String resultString = IOUtils.toString(fileHelper.getInputStream(resultFile),
+                                StandardCharsets.UTF_8);
+                        Map<String, String> data = summarizer.summarizeResults(resultString);
+                        tableRow.setData(data);
+
+                        // TODO: Load jsonSchema from metadata.json file and validate against schema -nbrown 11/2/23
+                        //File metadataJsonFile = extractFileFromZip(tempDir, downloadedFile, "metadata.json");
+                    } else {
+                        // TODO: Do we want more logging? -nbrown 11/2/23
+                        LOG.warn("Unable to load result file");
```
DwayneJengSage, line 621: We should include app ID, record ID, and assessmentGuid in the log message.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -545,4 +654,20 @@ private String getCalendarDateForRecord(HealthDataRecordEx3 record) {
...
+    protected File extractFileFromZip(java.io.File directory, java.io.File zipFile, String fileName) {
+        try {
+            ZipFile zFile = new ZipFile(zipFile);
+            zFile.extractFile(fileName, directory.getPath());
+            return fileHelper.newFile(directory, fileName);
+        } catch (ZipException e) {
+            LOG.error("Error extracting file from zip", e);
+        }
+        return null;
+    }
```
DwayneJengSage, line 668: TODO: Since we're unzipping user input, we need to defend against Zip Bomb attacks.

Also, the ZipFile library doesn't seem to play well with our InMemoryFileHelper, so we may need to refactor this.

src/main/java/org/sagebionetworks/bridge/exporter3/results/ArcResultSummarizer.kt
```diff
@@ -0,0 +1,30 @@
+package org.sagebionetworks.bridge.exporter3.results
+
+import org.sagebionetworks.bridge.rest.model.Assessment
+
+class ArcResultSummarizer(val assessment: Assessment) : AssessmentSummarizer {
+
+    init {
+        assert(assessment.frameworkIdentifier == FRAMEWORK_IDENTIFIER)
```
DwayneJengSage, line 8: What happens if this assert fails? Does it throw a RuntimeException? Should we instead throw a PollSqsWorkerBadRequestException so that SQS doesn't try to serve the request again?

src/main/java/org/sagebionetworks/bridge/exporter3/results/AssessmentResultSummarizer.kt
```diff
@@ -0,0 +1,70 @@
...
+     * representation of the result.
+     */
+    override fun summarizeResults(resultJson: String): Map<String, String> {
+        val assessmentResult: AssessmentResult = Serialization.JsonCoder.default.decodeFromString(resultJson)
+        val answers = assessmentResult.toFlatAnswers()
+        val columnNames = getColumnNames()
+        for (column in answers.keys) {
+            if (!columnNames.contains(column)) {
+                // TODO: How should we best log any unexpected columns? -nbrown 11/2/23
+                LOG.debug("Unexpected column: " + column + " when summarizing results for assessment: " + assessment.identifier)
```
DwayneJengSage, line 37: We should log this either as an info or a warning. Debug doesn't show up in our logs. Info we can make dashboards for, but there are no alarms. Warning will alarm if there are more than 100 per hour.

We should also include the app ID and record ID so we can trace the data through the logs.

DwayneJengSage (changes requested):

PR #180, opened 2023-11-20
DHP-1072: Add JSON to table row to exporter3 

DwayneJengSage:
https://sagebionetworks.jira.com/browse/DHP-1072

Replaces https://github.com/Sage-Bionetworks/BridgeWorkerPlatform/pull/178


src/main/java/org/sagebionetworks/bridge/exporter3/results/AssessmentResultSummarizer.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    init {
+        assert(assessment.frameworkIdentifier == FRAMEWORK_IDENTIFIER)
+    }
+
+    override val resultFilename: String
+        get() = "assessmentResult.json"
+
+    override fun canSummarize(assessment: Assessment): Boolean {
+        return assessment.frameworkIdentifier == "health.bridgedigital.assessment"
```
nategbrown9:
```suggestion
        return assessment.frameworkIdentifier == FRAMEWORK_IDENTIFIER
```
This was probably my fault.

nategbrown9 (approved):
Looks good.

DwayneJengSage:
@nategbrown9 Pull request is updated. If you have time to look over the incremental changes, either on Wednesday or Monday morning. Thanks!

Also, minor updates to https://github.com/Sage-Bionetworks/BridgeWorkerIntegrationTests/pull/63 Thanks!

nategbrown9 (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +600,127 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+            tableRow = new UploadTableRow();
+            tableRow.setAssessmentGuid(assessmentGuid);
+            tableRow.setCreatedOn(upload.getRequestedOn());
+            tableRow.setRecordId(recordId);
+            tableRow.setHealthCode(record.getHealthCode());
+            tableRow.setParticipantVersion(record.getParticipantVersion());
+
+            // Metadata.
+            Map<String, String> tableRowMetadata = new HashMap<>();
+            tableRowMetadata.put(METADATA_KEY_CLIENT_INFO, record.getClientInfo());
```
syoung-smallwisdom, line 623:
```suggestion
            // TODO: syoung 11/22/2023 Get the client info and user agent from the upload metadata rather than the participant record.
            tableRowMetadata.put(METADATA_KEY_CLIENT_INFO, record.getClientInfo());
```

Ideally, this would be set from the OS version, device, etc from when it was uploaded rather than the last call from the device to the server.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +600,127 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+                    // We still want to include a blank value in the table row, so that the column is
+                    // present in the CSV.
+                    tableRowMetadata.put(key, "");
+                }
+            }
+            tableRow.setMetadata(tableRowMetadata);
+
+            // Get data groups from participant. Sort the data groups so that the order is deterministic. (Null means
+            // it sorts using the natural ordering, ie alphabetical order.)
+            List<String> dataGroupList = new ArrayList<>(participant.getDataGroups());
```
syoung-smallwisdom, line 639: Again, this use to be included in the "metadata.json" file b/c it could change. While in practice, the newer apps don't change data groups based on surveys, etc. Those groups *can* change. ¯\_(ツ)_/¯

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +600,127 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+            dataGroupList.sort(null);
+            tableRow.setTestData(dataGroupList.contains(DATA_GROUP_TEST_USER));
+            tableRowMetadata.put(METADATA_KEY_DATA_GROUPS, Constants.COMMA_JOINER.join(dataGroupList));
+
+            // Open the zip file only if we recognize the assessment.
+            Assessment assessment = bridgeHelper.getAssessmentByGuid(appId, assessmentGuid);
+            AssessmentConfig assessmentConfig = bridgeHelper.getAssessmentConfigByGuid(appId, assessmentGuid);
```
syoung-smallwisdom: If there is no config will this throw an exception? (I don't remember, sorry)

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
```diff
@@ -528,6 +600,127 @@ private String getRawS3KeyForUpload(String appId, String studyId, Upload upload,
...
+                                tableRowMetadata.put(key, "");
+                            }
+                        }
+                    } else {
+                        LOG.warn("Unable to load metadata.json file for appId=" + appId + ", recordId=" + recordId +
+                                ", assessmentGuid=" + assessmentGuid);
+                    }
+
+                    // Summarize results file.
+                    File resultFile = unzippedFileMap.get(summarizer.getResultFilename());
```
syoung-smallwisdom, line 701: FWIW: This may need to be refactored in the future to accommodate assessments that use multiple files - voice analysis, images, motion sensors, etc.

src/main/java/org/sagebionetworks/bridge/exporter3/results/ArcResultSummarizer.kt
```diff
@@ -0,0 +1,29 @@
+package org.sagebionetworks.bridge.exporter3.results
```
syoung-smallwisdom, line 1: I was so confused as to why I suddenly could read the code fluently. ;)

syoung-smallwisdom (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
DwayneJengSage: We do both. userAgent and clientInfo comes from the User-Agent header when the upload is first created, so it won't get overwritten during redrives. It also doesn't depend on the file format, so it's available for every assessment.

deviceInfo comes from metadata.json and is only available in assessments that we recognize.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
DwayneJengSage: We decided in a design meeting earlier this month that data groups don't change often enough to make it worth addressing this ambiguity.

Long-term, if we need to address this ambiguity, we would need a flag during CSV generation that specifies whether we use the historical participant information or the most up-to-date participant information. Over the past several years, researchers have asked for both, so we would need to support both.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
DwayneJengSage: Good catch. This will in fact throw. I'll add try-catches.

src/main/java/org/sagebionetworks/bridge/exporter3/Exporter3WorkerProcessor.java
DwayneJengSage: Good idea. In the future, we might need to pass in the whole file map into the Summarizer. I'll add a comment.

PR #181, opened 2023-12-07
DHP-1160 Fix No CSVs for assessments without a summarizer

DwayneJengSage:
https://sagebionetworks.jira.com/browse/DHP-1160

When there's no summarizer, there's no data. When we try to copy this data into the study-specific table row, we get a Null Pointer Exception.

While fixing this bug, I found another bug where UploadTableEnabled is null.

I edited/added unit tests to verify the bugs, then fixed the bugs and re-ran the unit tests to verify the fix.


nategbrown9 (approved):
Looks good.

syoung-smallwisdom (approved):
looks good

PR #183, opened 2024-01-08
Add implementation to ArcResultSummarizer

nategbrown9:
This is now ready to merge

- [x] ArcAssessmentsKMP published to maven repo
- [x] Tests
- [x] Create DIAN container assessment


src/main/java/org/sagebionetworks/bridge/exporter3/results/ArcResultSummarizer.kt
```diff
@@ -1,6 +1,8 @@
 package org.sagebionetworks.bridge.exporter3.results
 
+import edu.wustl.arc.result.*
```
DwayneJengSage: Our code convention is that we list out all imported packages instead of using wild cards. There should be a setting to disable this in IntelliJ (or at least to set it to something arbitrarily high like 999).

DwayneJengSage (approved):
Overall looks good. One non-blocking change about imports and code conventions.

I think it would be worth adding an integ test for the Arc summarizer, just to make sure everything is wired together correctly and behaving as expected. See https://github.com/Sage-Bionetworks/BridgeWorkerIntegrationTests/blob/develop/src/test/java/org/sagebionetworks/bridge/exporter/integration/Exporter3Test.java#L702 for how we did the integ test for Surveys. (Might need a little bit of finagling.)

src/main/java/org/sagebionetworks/bridge/exporter3/results/ArcResultSummarizer.kt
nategbrown9: I had updated the setting for Java files, but there is a separate one for Kotlin.

PR #184, opened 2024-02-13
DHP-1154 Adherence Report Bug

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/DHP-1154

getAllApps() sets summary=true, which only includes app name and ID and not the flag that enables adherence report. summary=false can only be called by Superadmin.

This change is a slight refactor so that we call getApp() for every app ID, even in the case of all apps, so we can get all app properties including the flag that enables adherence report.

(Unit tests didn't need to be changed, since they were already mocking getApp() anyway.)


Erin-Mounts (approved):
Looks good to me

nategbrown9 (approved):
Looks good.

PR #187, opened 2024-03-05
tweak log message for CSV Worker

DwayneJengSage:
Dashboard currently only shows the study ID, which isn't very descriptive. Add both the study ID and the study name.


Erin-Mounts (approved):
looks good

nategbrown9 (approved):

